Issue #13, opened 2013-06-18
robot_status.h lists fields in the wrong order


gavanderhoorn:
The header lists them in a different order than the respective `::load(..)` and `::unload(..)` methods use. This is rather confusing, as other headers list the fields in the same order as those methods.

gavanderhoorn:
This has been merged into `hydro-devel` and `groovy-devel`.

PR #14, opened 2013-06-20
Fix issue #13 (field order in docs of RobotStatus msg)

gavanderhoorn:
As per subject.


gavanderhoorn:
Opened against `master`, should also be merged with `groovy-devel` and `hydro-devel`.

JeremyZoss:
Fixed in 4405b0d, cb23c64.

gavanderhoorn:
It seems `hydro-devel` is missing this 'fix'. Are you going to merge it in there as well?

JeremyZoss:
Sure.

That branch didn’t exist when I made the first merge (I don’t think).
- Jeremy

gavanderhoorn:
@JeremyZoss: ok, but it will need the fix :)

PR #16, opened 2013-07-07
Fix issue #15 (missing const modifier)

gavanderhoorn:
Small change.


gavanderhoorn:
Again, should also be merged with `groovy-devel`.

PR #30, opened 2013-08-02
Fix issue #29: Log messages that display errno should display the associated error text instead.

### LINKED ISSUE #29: Log messages that display errno should display the associated error text instead.
Several (i.e. many) users have had issues with socket connection errors to different controllers.  When this occurs, the log messages contain the last error code stored in errno.  To determine the root cause of the fault, the error code value must be looked up.  A better log message should be used the utilizes the strerror() function ( http://linux.die.net/man/3/strerror ).  This function returns a string description of error code (presumably similar to comments in error.h).

gavanderhoorn:
Add error message to socket errors (instead of just errno).

Changes `socketError(..)` output from:

> Failed to connect to server, rc: -1, errno: 113

to:

> Failed to connect to server, rc: -1. Error: 'No route to host' (errno: 113)

which should give some more information to the user.

Fix #29.

### DIFF
--- a/simple_message/include/simple_message/socket/simple_socket.h
+++ b/simple_message/include/simple_message/socket/simple_socket.h
@@ -203,7 +203,8 @@ protected:
 
   void logSocketError(const char* msg, int rc)
   {
-    LOG_ERROR("%s, rc: %d, errno: %d", msg, rc, errno);
+    int errno_ = errno;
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Hm, just realised this should probably take differences between the compile targets into account (ie: controllers and pc). Will look at that later.

PR #32, opened 2013-08-07
Joint feedback handler

liborw:
This is an initial implementation of the JointFeedbackHandler. In this state the handler is able to handle JointFeedbackMessage, assign joint names based on the robot id and publish it /joint_state and /joint_feedback message.

### DIFF
--- a/industrial_robot_client/CMakeLists.txt
+++ b/industrial_robot_client/CMakeLists.txt
@@ -13,7 +13,8 @@ add_definitions(-DROS=1)  #build using ROS libraries
 add_definitions(-DLINUXSOCKETS=1)  #build using LINUX SOCKETS libraries
 
 set(SRC_FILES src/joint_relay_handler.cpp
-	      src/robot_status_relay_handler.cpp
+              src/joint_feedback_handler.cpp
+              src/robot_status_relay_handler.cpp
               src/joint_trajectory_downloader.cpp
               src/joint_trajectory_streamer.cpp
               src/joint_trajectory_interface.cpp
@@ -85,6 +86,7 @@ include_directories(include
 
 add_library(industrial_robot_client ${SRC_FILES})
 target_link_libraries(industrial_robot_client simple_message)
+target_link_libraries(industrial_robot_client industrial_utils)
 
 add_library(industrial_robot_client_bswap ${SRC_FILES})
 target_link_libraries(industrial_robot_client_bswap simple_message_bswap)
--- a/industrial_robot_client/include/industrial_robot_client/joint_feedback_handler.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_feedback_handler.h
@@ -1,0 +1,144 @@
+/*
+* Software License Agreement (BSD License) 
+*
+* Copyright (c) 2011, Southwest Research Institute
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*   notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above copyright
+*   notice, this list of conditions and the following disclaimer in the
+*   documentation and/or other materials provided with the distribution.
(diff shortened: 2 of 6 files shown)


JeremyZoss:
This implementation loses most of the [modular design](http://www.ros.org/wiki/industrial_robot_client/design#JointRelayHandler) present in the current [joint_relay_handler](https://github.com/ros-industrial/industrial_core/blob/groovy-devel/industrial_robot_client/include/industrial_robot_client/joint_relay_handler.h).  I think that a better solution would be to adapt the current joint_relay_handler architecture to be less dependent on the specific message type used for robot communication.

The design used in the FS100 driver allows for both JOINT_FEEDBACK and JOINT_STATE messages to be handled by the same set of methods.  See [here](https://github.com/ros-industrial/motoman/blob/groovy-devel/fs100/src/industrial_robot_client/joint_feedback_relay_handler.cpp) for my suggested implementation.  In this design, you just re-implement convert_message() to convert between any new type of simple_message to a generic ROS JointTrajectoryPoint for feedback.

liborw:
The main purpose of this implementation was the support the robot_id variable and dual-arm robots. That is (I think) the main advantage of JOINT_FEEDBACK over the JOINT_STATE message. The FS100 design doesn't take into account multiple control groups.

JeremyZoss:
The FS100 code was intended to use the "robot_id" field to handle multiple control groups.  In this scheme, the controller will send a separate JOINT_FEEDBACK message for each control group.  A separate joint_feedback_relay_handler would be created for each control group, specifying the robot_id and controller_joint_names appropriate to that motion group.  The existing FS100 code should then listen only to JOINT_FEEDBACK messages corresponding to the specified robot_id and broadcast the state data appropriately.

This hasn't been thoroughly tested, but that was the design intent.

liborw:
Actually, I assumed this intention and tried.

The problem is that the robot_id and controller_joint_names can be set to entire node not to particular joint_feedback_relay_handler, but that could be probably changed. The other thing that doesn't seem right to me is that each joint_feedback_relay_handler is called for each message. The control group manager and handlers would be more appropriate in this concept.

JeremyZoss:
You should be able to set robot_id and controller_joint_names either on a per-node or per-handler basis.  For a per-node basis, set the ROS parameters "robot_id" and "controller_joint_names".  For a per-handler basis, the current code allows you to specify these values in the handler constructors:

```
joint_feedback_relay_handler r1_handler(1), r2_handler(2);
r1_handler.init(conn, r1_joint_names);
r2_handler.init(conn, r2_joint_names);
```

This may break down when you try and add multiple message handlers for the same type to the MessageManager.  That may not be allowed.

And, I agree that it's a little "messy".  I would welcome any suggestions for how to clean up multi-robot support while still maintaining a reasonably flexible/extensible design.  Perhaps a better approach would be to allow you to specify a map of robot_id to joint-name-list, so that multiple robots could be handled by the same handler?

```
joint_feedback_relay_handler h;
map<int, vector<string>> joint_names;
joint_names[1] = ["left_joint_1", "left_joint_2", ...]
joint_names[2] = ["right_joint_1", "right_joint_2", ...]
h.init(joint_names)
```

Ideally, this could also be specified via ROS param.  It's a little messier than the current robot_id (int) and controller_joint_names (vector-of-strings), but could probably be done.  It might be nice to keep the configuration of this node and the motion-control node consistent, to avoid forcing lots of redundant configuration settings.

liborw:
The second approach is actually what I did in the joint_feedback_handler.

I have looked into the possibility of passing the map through ROS param, but the roscpp can only handle string keys  (see [roscpp: Parameter Server](http://www.ros.org/wiki/roscpp/Overview/Parameter%20Server)).

liborw:
What about implementing a "control group manager" that will process the JOINT_FEEDBACK (or possibly JOINT_POSITION) message to some intermediate structure (can be JointTrajectoryPoint even though it doesn't feel right) and call appropriate handler based on the robot id.

The handlers would be independent from the JOINT message type and also the message would be processed only once.

shaun-edwards:
This issue will be addressed by an REP.

Issue #25, opened 2013-07-25
Number of received bytes differ from message length (SmplMsgConnection::receiveMsg).


liborw:
When testing the motoman driver, I got this error:

```
$ rosrun fs100 robot_state
[ INFO] [1374677672.437211315]: Added message handler for message type: 0
[ INFO] [1374677672.440014448]: Robot state connecting to IP address: '192.168.0.101:50241'
[ WARN] [1374677672.440910442]: Unable to find user-specified joint names in 'controller_joint_names'
[ WARN] [1374677672.441766703]: Unable to find URDF joint names in 'robot_description'
[ INFO] [1374677672.441864812]: Using standard 6-DOF joint names: [joint_1, joint_2, joint_3, joint_4, joint_5, joint_6]
[ INFO] [1374677672.442550563]: Connected to server
[ INFO] [1374677672.442607446]: Initializing message manager with default comms fault handler
[ INFO] [1374677672.442657083]: Default communications fault handler successfully initialized
[ INFO] [1374677672.442685145]: Initializing message manager
[ INFO] [1374677672.442726762]: Added message handler for message type: 1
[ INFO] [1374677672.446381376]: Added message handler for message type: 15
[ INFO] [1374677672.447653822]: Entering message manager spin loop
[ERROR] [1374677672.495079875]: Get unload pointer failed, buffer size: 3, smaller than byte size: 4
[ERROR] [1374677672.495149280]: Unload pointer returned NULL
[ERROR] [1374677672.495206461]: Failed to unload message joint: 5 from data[3]
[ERROR] [1374677672.495241827]: Failed to unload joint feedback positions
[ERROR] [1374677672.495278751]: Failed to unload joint feedback message data
[ERROR] [1374677672.495314845]: Failed to initialize joint feedback message
[ERROR] [1374677672.495351389]: Failed to convert SimpleMessage
[ WARN] [1374677672.495409474]: Recieved zero bytes: 0
[ERROR] [1374677672.495445331]: Failed to initialize message
[ERROR] [1374677672.495485843]: Failed to receive incoming message
[ WARN] [1374677672.495522284]: Send failure, no callback support
^C[ INFO] [1374677674.162045284]: Connection failed, attempting reconnect
[ERROR] [1374677674.162110342]: Failed to connect to server, rc: -1, errno: 106
[ WARN] [1374677674.162132291]: Not connected, bytes not sent
[ERROR] [1374677674.162152122]: Failed to receive message length
[...]

PR #33, opened 2013-08-13
Issue #25 fix

shaun-edwards:
These changes address split packet issues associated with the tcp socket read command (issue #25). 

The read command was changed to utilize the socket select function to query the socket buffer state.  This allows the read to poll the socket for new data in a "interrupt-able" way.  This means that "Control-C" will break out of the program as expected.  The previous approach of simply reading or performing a blocking read did not seem to work this way.

Unit tests were performed that demonstrated the problem and verified the fix.

Minor changes were made to allow parallel test execution (default under catkin).

### DIFF
--- a/simple_message/CMakeLists.txt
+++ b/simple_message/CMakeLists.txt
@@ -85,6 +85,8 @@ include_directories(include
   ${catkin_INCLUDE_DIRS}
 )
 
+# NOTE: All test files require TEST_PORT_BASE to be defined.  Defining different
+# ports for each test executable allows them to run in parallel.
 
 # DEFAULT LIBRARY (SAME ENDIAN)
 add_library(simple_message ${SRC_FILES})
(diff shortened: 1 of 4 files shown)


JeremyZoss:
I may need you to help talk me through this. A few questions:

1) There does not seem to be a way for the user to specify an overall timeout when calling receiveBytes(). I'm not clear on what benefit the existing timeout provides. Why not just block on select() forever? Is this timeout what allows you to handle Ctrl-C signals? If so, it should be documented in comments, since it's not clear.
2) When receiveBytes() called and data not ready (e.g. while waiting for a command), code will log a WARN every time the socket times out. This could potentially generate continuous WARN msgs at 1Hz or so. Should downgrade this to a DEBUG msg?
3) Why are there three separate "error" returns from poll() that seem to be handled independently? Wouldn't the old isReadyReceive() function be adequate?
4) You don't seem to have a test case for: sendBytes(ONE_INT) receive bytes(TWO_INTS). This will probably block forever in the current design.

It seems like you've addressed split-packet concerns (repeated READ calls until expected # of bytes received), but there is still no way for high-level users to specify an overall timeout on a read. To higher-level code, it will still appear as if the code is blocking "forever" until the requested data has been read. This is okay, but we should log a new enhancement issue to track this need.

Ideally, I would like to see this fix include the API changes required to implement a true read-timeout, specified by the user during a call to receiveBytes(). This could be propagated through the existing code base using parameters that default to wait-forever, so as to not break existing code. There may be some higher-level concerns, though, as I don't think the simple_message protocol will gracefully handle the case where only part of a message is read. It will probably require the user to drop/re-establish connection to reset the simple_message data stream.

shaun-edwards:
@JeremyZoss, I had a long response typed out, but then github lost it.  So in short, I made some modifications to address your concerns.  As far as a read timeout, I think we should submit a new issue to address that.

Issue #50, opened 2013-12-27
planning_request_adapters: missing install rule for plugin xml (hydro)


gavanderhoorn:
Seems the `CMakeLists.txt` is missing an install rule for the `planning_request_adapters_plugin_description.xml` file in the `hydro-devel` branch.

gavanderhoorn:
Fixed in 30dfc20f (hydro-devel).

PR #51, opened 2013-12-27
Fix issue #50: traj_filters: adds install rule to CMakeLists.txt for plugin info xml

gavanderhoorn:
As per subject.


gavanderhoorn:
@jrgnicho, @shaun-edwards: review?

shaun-edwards:
+1

Issue #53, opened 2013-12-28
robot_simulator: hard coded Python path


gavanderhoorn:
See ros-industrial/ros_industrial_issues#9.

PR #54, opened 2013-12-28
Fix issue #53: hard coded Python path in robot_simulator script

gavanderhoorn:
As per subject. Minor change. Perhaps we should explicitly specify python2 in the shebang.

This is against `groovy-devel`.


shaun-edwards:
+1, please submit a pull request for the hydro-devel branch as well.

Issue #61, opened 2014-01-07
Industrial deprecated package should updated


shaun-edwards:
The industrial deprecated package is meant to ease deprecation of packages within ROS and ROS-Industrial.  It should be cleaned of anything that is officially unsupported (these are mostly arm-navigation support) and populated with any new deprecated items (none that I know of).

PR #64, opened 2014-01-08
Remove deprecates items from industrial depricated

shaun-edwards:
Fix for issue #61.


shaun-edwards:
@jrgnicho, please review.

industrial_deprecated/CMakeLists.txt
```diff
@@ -1,68 +1,10 @@
...
-# )
-
-## Generate added messages and services with any dependencies listed here
-# generate_messages(
-#   DEPENDENCIES
-#   std_msgs
-# )
+project(industrial_deprecated)
 
+find_package(catkin REQUIRED COMPONENTS)
```
gavanderhoorn, line 5: Does this work like that, with the empty `REQUIRED COMPONENTS` argument?

`find_package(catkin)` is what I mostly see when there are no actual required dependencies.

shaun-edwards, line 5: It works, but certainly not required.  I kept it in since I expect to add some dependencies in the future (as a reminder that they have added there as well).

Issue #67, opened 2014-01-21
robot_client: dependent pkgs fail to link (catkin_make_isolated)


gavanderhoorn:
Identical to ros-industrial/industrial_core#46. See ros-industrial/fanuc#56 for some more info.

Issue #46, opened 2013-12-04
industrial_robot_client: build failure when using catkin_make_isolated


gavanderhoorn:
As per this email to the ros-industrial mailing list: [Compiling ros-industrial with catkin_make_isolated](https://groups.google.com/d/msg/swri-ros-pkg-dev/wwcyZD5bVtw/7t-R6kDbkNIJ).

Building `industrial_robot_client` using `catkin_make_isolated` fails with the following error:

```
==> make -j1 in '/home/user/ros/industrial_groovy_isolated/build_isolated/industrial_robot_client'
Linking CXX shared library /home/user/ros/industrial_groovy_isolated/devel_isolated/industrial_robot_client/lib/libindustrial_robot_client.so
/usr/bin/ld: cannot find -lsimple_message
collect2: ld returned 1 exit status
make[2]: *** [/home/user/ros/industrial_groovy_isolated/devel_isolated/industrial_robot_client/lib/libindustrial_robot_client.so] Error 1
make[1]: *** [CMakeFiles/industrial_robot_client.dir/all] Error 2
make: *** [all] Error 2
<== Failed to process package 'industrial_robot_client': 
  Command '/home/user/ros/industrial_groovy_isolated/devel_isolated/simple_message/env.sh make -j1' returned non-zero exit status 2
```

This seems to be due to the fact that CMake cannot resolve the reference in `target_link_libraries(..)` to the target `simple_message` (which is defined in the `simple_message` pkg), and tries to link with `libsimple_message` (no full path).

In a 'normal' `catkin_make` invocation, all targets share a single namespace, hence CMake can resolve the reference. With `catkin_make_isolated` all pkgs are build in isolation, the target `simple_message` doesn't exist in the namespace for `industrial_robot_client` and the above error is reported.

gavanderhoorn:
The root cause for this seems to be the fact that `simple_message` does not export its libraries in the `catkin_package(..)` call. Adding `LIBRARIES simple_message` to the `catkin_package(..)` statement allows `industrial_robot_client` to be build again (but this is obviously _not_ a solution).

One possible solution would be to find a better solution to the issue described in #10.

shaun-edwards:
@gavanderhoorn, I'm not sure I have a proper solution to this problem.  Do you have anything in mind.  If not, I was going to ask the rosbuild list for their input.

gavanderhoorn:
Well, off the top of  my head I can come up with two approaches:
[...]

PR #69, opened 2014-01-21
Fix issue #67: update workaround for #46 to include industrial_robot_client

gavanderhoorn:
Also removes the need to invoke the `X_linker_setup()` function in dependent packages.


shaun-edwards:
+1

PR #70, opened 2014-02-17
Fixing receiveBytes for UDP

kphawkins:
Changes to this method meant that when a call to receiveBytes
with num_bytes = 0 meant that the function exited without trying
to read.  Checks were added to make sure, in the case num_bytes = 0
to call at least once.


shaun-edwards:
@kphawkins - I looked through the `SimpleSocket` code to determine how and why this was required.  From `UdpSocket::rawReceiveBytes`, it looks like you are ignoring the number of bytes and just reading the maximum.  I'd like to understand why this slight of hand is required and why the expected behavior which would be to read the appropriate number of bytes does not work for UDP.  Because UDP was not implemented at the same time as TCP, I wouldn't be surprised if the socked abstraction isn't quite correct.

kphawkins:
See https://github.com/ros-industrial/industrial_core/blob/hydro-devel/simple_message/src/socket/udp_socket.cpp#L75 . I don't have enough intuition here to know if this is the appropriate way of doing UDP sockets, but perhaps you see a bug here?

Thanks for looking at this.

kphawkins:
It seems like the typical message reading was overridden at some point. Have your changes fixed this problem? https://github.com/ros-industrial/industrial_core/blob/hydro-devel/simple_message/include/simple_message/socket/udp_socket.h#L68-73

shaun-edwards:
Ok, looks like I am the one to blame here ;).  The comment in `udp_socket.h` is still true, see [here](http://stackoverflow.com/questions/675812/successive-calls-to-recvfrom-loses-data).  The UDP code appears to read all the data in a single read (as opposed to tcp socket calls which can read the `length` portion of the message and the `data` portion separately).  The `UdpSocket::receiveMsg` assumes that only a single message is read at a time.  If more is read, this data is lost (it would have been lost anyway because the socket read would have lost it).  The underlying assumption is that if the socket is read fast enough, then only a single message will be in the buffer.

A better approach than overriding `UdpSocket::receiveMsg` might be to make `UdpSocket::rawReceiveBytes` act like a tcp socket read.  In order to do this, we could utilize `this->buffer_` as a temporary buffer.  This would add some book-keeping to the UDP socket and care would have to be taken not to overrun the buffer, but I think this is manageable.

kphawkins:
Can I merge this commit?

shaun-edwards:
I still believe the better solution is to fix the underlying UDPSocket library.  I'm open to merging this if we do the following:
- Open an issue to fix the problem as I suggest (this merge would only be a temporary solution)
- Add a unit test that demonstrates the problem and verifies the fix.  This way, we guard against future changes that break UDP.

kphawkins:
Okay, here's a more formal fix for UDP communication.

This should now make UDP sockets act almost exactly like the
TCP sockets.  The primary difference I notice at the moment
is that makeConnect is necessarily blocking for UDP, while for
TCP it just connects without the need for a handshake.

Here is a list of the changes:
- UdpSocket now has its own read buffer. For rawReceiveBytes,
   it now only does a RECV_FROM to its UDP-specific buffer when there
   is no data in it.  Then it copies whatever it can from that buffer into the
   requested buffer.  The rawPoll operation acts almost the same as the previous poll,
   except it goes ahead and returns as ready when there is data left in the internal
   buffer. Thus, we should get almost the same functionality overall as the TCP
   code.
- Removed original "fixes" I put in. receiveBytes acts the same as originally.
- SimpleSocket::setConnected is now public (this is the only way you can
   encourage a socket to try and connect again, if you timeout in
   cyclic communication, the server should set this to false and
   try connecting again when it gets information on the point)
- SimpleSocket::poll is now a virtual method named rawPoll to be implemented by *Socket
   (this is due to how UDP does reads now).  TCP's poll code is identical to the original.
- UdpServer::makeConnect doesn't hard-block waiting for the handshake message anymore.
- I have refactored most of the utest.cpp code to make it possible to run the same
   tests for both TCP and UDP.  Most of the code is identical, except that the makeConnect
   calls have to be run differently. Since for UDP, each call is blocking until the
   handshake is completed, you must call one of them in a thread.  It seems that this
   works fine for both TCP and UDP though.
- I have added a new test which enables the UDP_TEST macro in the utest.cpp.

gavanderhoorn:
@kphawkins: personally, I like this way better than the previously proposed changes. One question:

> - UdpSocket now has its own read buffer. For rawReceiveBytes, it now only does a RECV_FROM to its UDP-specific buffer when there is no data in it.  Then it copies whatever it can from that buffer into the requested buffer. [..]

Why only read from the socket when the internal read buffer of `UdpSocket` is empty? Any reason not to make that internal buffer a bit larger (say 8KB) and just always transfer any available bytes from socket buffer to the internal one?

kphawkins:
You should only hit RECV_FROM after you're sure there is data in the
socket, or else you're blocking in kernel space. When you read in a packet,
it reads in as much as the regular buffer can handle, but doesn't seem to
combine multiple packets into one. To do this, you'd have to do some sort
of loop which is precisely what readBytes does, which keeps checking for
buffer data after every read.

shaun-edwards:
@kphawkins, thanks for coming back around to this one.  I really appreciate it.

> SimpleSocket::setConnected is now public (this is the only way you can
> encourage a socket to try and connect again, if you timeout in
> cyclic communication, the server should set this to false and
> try connecting again when it gets information on the point)

Can you give more details on why this is required.  Everything (servers & clients) inherit from this class, so they should have access to the protected method.  I could not find where this method was called in a "public" way.  I reverted this method to protected and all of the unit tests still pass.

> I have refactored most of the utest.cpp code to make it possible to run the same
> tests for both TCP and UDP. Most of the code is identical, except that the makeConnect
> calls have to be run differently. Since for UDP, each call is blocking until the
> handshake is completed, you must call one of them in a thread. It seems that this
> works fine for both TCP and UDP though.

I appreciate you updating the unit tests (this should ensure we don't break things again).  Just an FYI (and more of a critique of my original unit test implementation), Typed-tests are the better way to achieve this interface level testing (see [here](https://code.google.com/p/googletest/wiki/AdvancedGuide#Typed_Tests)).  No need to address this.

shaun-edwards:
ping....@kphawkins...can you address my question above?

kphawkins:
Usecase: If an implementing server expects cyclic messages and gets no
response from the client for several cycles, it needs some way of
disconnecting. After setting the interface to false, the server can begin
the connection procedure again. This allows a client to properly reconnect.

kphawkins:
I have no unit test for this particular behavior.

shaun-edwards:
I apologize for being difficult.  Doesn't an implementing server inherit form the socket base class and thus have access to the protected method?

kphawkins:
Use case:
https://github.com/gt-ros-pkg/universal_robot/blob/hydro-c-api-ur10-moveit/ur_ctrl_server/src/ur_controller_iface.cpp#L26-L52

This method gets called cyclically.
makeConnect does nothing unless the connected state is false.
setConnected(false) called at line 51.

shaun-edwards:
OK...now I get it.  I can see where the user of the connection may want to close/reset the connection for whatever reason.  Because the `setConnected` method just sets an internal variable, I would rather create a new function (`reset` or `disconnect`) that could do additional cleanup (even if all it does is call `setConnected` for now).  Please add this new function to the PR.

On a side note (and not required for this PR), I have always envisioned these socket communications as maintaining their connection state internally.  UDP doesn't do this at the protocol level, so the UDP class would have to do this.  One idea is to monitor the last time a message was received.  If it has been too long, and `isConnected` is called, a [handshake](https://github.com/ros-industrial/industrial_core/blob/hydro-devel/simple_message/src/socket/udp_server.cpp#L125-L143) could be tried.  If the handshake failed, then the socket would be set to disconnected.

kphawkins:
The only way you can do this is when you have cyclic communication, or when
you have a reasonable estimate of when a message should appear. Right now,
there is no guarantee that any data will be sent cyclically by the server.

You could potentially setup a cyclic ping which is automatically run, but
you'd have no guarantee that the server would call a method which could
send it off.

You could designate a timeout for a particular message channel, such that
if the client doesn't get messages for X seconds then the client should
disconnect. I think that is more complicated though.

I agree that having a setDisconnected method makes sense, there's no reason
for the user to falsely connect.

shaun-edwards:
ping?  Are you going to be able to make the requested changes?

kphawkins:
I'll try to add them in the coming week.

kphawkins:
Alright, will this work?  I looked into making unit tests, but the use case is somewhat complicated.

shaun-edwards:
Totally missed your response.  I apologize @kphawkins.  Thanks very much for the contribution!   +1

PR #71, opened 2014-02-23
Fix compilation on Mac OS X 10.9

marcoesposito1988:


industrial_robot_client/include/industrial_robot_client/joint_trajectory_interface.h
```diff
@@ -104,9 +104,19 @@ class JointTrajectoryInterface
    * \param velocity_limits map of maximum velocities for each joint
    *   - leave empty to lookup from URDF
    * \return true on success, false otherwise (an invalid message type)
-   */
+   */
+#ifndef __clang__
```
shaun-edwards, line 108: Is it the default argument that is giving LLVM trouble?  Is this something we should avoid to make this package more cross platform friendly?

shaun-edwards:
I'd like to understand the issue a little better (see my comments above).  If there is an issue with default parameters perhaps we should curtail their use.  In fact the ROS/Google C++ standard is not to use them(see [here](http://google-styleguide.googlecode.com/svn/trunk/cppguide.xml#Default_Arguments))

marcoesposito1988:
Yes, even though this was already fixed, looks like this is still a problem for the latest Xcode to date (see http://llvm.org/bugs/show_bug.cgi?id=13657 )

So I'd say that either this pattern should be avoided, or this ifdef could be left here, so that it will be clear that it can be removed as soon as it doesn't create any more problem.

gavanderhoorn:
@shaun-edwards wrote:

> I'd like to understand the issue a little better (see my comments above). If there is an issue with default parameters perhaps we should curtail their use. In fact the ROS/Google C++ standard is not to use them(see [here](http://google-styleguide.googlecode.com/svn/trunk/cppguide.xml#Default_Arguments))

I vote for just avoiding default arguments completely - as recommended in the linked style guide. The proposed alternative - method overloading - _is_ support everywhere, and should not add too much methods.

It would also allow us to avoid adding `#ifdefs` everywhere, which I think don't improve the readability of the code.

gavanderhoorn:
As to the linker issue: @shaun-edwards: perhaps we should merge some of the changes to the build scripts from `hydro-devel` into `groovy-devel`? Leaving out the release related commits, it would seem there haven't been that many (see: [groovy-devel...hydro-devel](https://github.com/ros-industrial/industrial_core/compare/groovy-devel...hydro-devel)).

shaun-edwards:
I was just bitten by default arguments on another project (specifically because the "default" changed underneath me).  I think this is just one more reason not to use them.  

@marcoesposito1988, would you be willing to modify your pull request to remove the default arguments and instead use function overloading (at least for the parts that are causing you issues)

shaun-edwards:
@gavanderhoorn, can  you start another issue to track merging back to groovy-devel.  I'd like to capture the discussion but not pollute this issue.

shaun-edwards:
I'm closing this PR due to a lack of activity.

PR #83, opened 2014-08-25
Trajectory action should abort on robot error.

hersh:
When a robot status message is received with a non-zero error code, any active trajectory goal should be aborted.  Without this, the only way to know that a failure has happened is for the action client to time-out.

Also fixed some compiler warnings about printing unsigned integers with %d format.

### DIFF
--- a/industrial_robot_client/src/joint_trajectory_action.cpp
+++ b/industrial_robot_client/src/joint_trajectory_action.cpp
@@ -73,6 +73,16 @@ JointTrajectoryAction::~JointTrajectoryAction()
 void JointTrajectoryAction::robotStatusCB(const industrial_msgs::RobotStatusConstPtr &msg)
 {
   last_robot_status_ = msg; //caching robot status for later use.
+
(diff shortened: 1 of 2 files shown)


shaun-edwards:
@hersh, thanks for this contribution.  I'm going to try it out on our hardware before I accept the PR.  Should be a couple of days.

gavanderhoorn:
I like this, but should the abort of the goal (and sending of the `FollowJointTrajectoryResult`) perhaps be done somewhere else instead of directly here in the RobotStatus callback? Maybe in `controllerStateCB(..)`?

gavanderhoorn:
Also: perhaps we should actually check the `in_error` field, instead of `error_code` (or both)? I can imagine not all errors (on the industrial controller) immediately lead to an inability to complete the goal?

gavanderhoorn:
Rebase + improvements in #271.

really late, but I'd still like to thank @hersh for making the effort to submit a PR.

PR #84, opened 2014-10-14
Silent warnings

VictorLamoine:
**EDIT:** Change `joints`/`links` print format specifier to avoid compilation warnings:

```
/home/dell/ros_fanuc_ws/src/industrial_core/industrial_utils/src/utils.cpp: In function ‘bool industrial_utils::findChainJointNames(const boost::shared_ptr<const urdf::Link>&, bool, std::vector<std::basic_string<char> >&)’:
/home/dell/ros_fanuc_ws/src/industrial_core/industrial_utils/src/utils.cpp:81:1117: warning: format ‘%d’ expects argument of type ‘int’, but argument 8 has type ‘std::vector<boost::shared_ptr<urdf::Joint> >::size_type {aka long unsigned int}’ [-Wformat=]
   ROS_DEBUG("Found %d child joints:", joints.size());                                                                                                                                                                                                                                                                                                                                                                         ^
/home/dell/ros_fanuc_ws/src/industrial_core/industrial_utils/src/utils.cpp:103:1116: warning: format ‘%d’ expects argument of type ‘int’, but argument 8 has type ‘std::vector<boost::shared_ptr<urdf::Link> >::size_type {aka long unsigned int}’ [-Wformat=]
   ROS_DEBUG("Found %d child links:", links.size());

```

### DIFF
--- a/industrial_utils/src/utils.cpp
+++ b/industrial_utils/src/utils.cpp
@@ -78,7 +78,7 @@ bool findChainJointNames(const boost::shared_ptr<const urdf::Link> &link, bool i
 
   // check for joints directly connected to this link
   const joint_list &joints = link->child_joints;
-  ROS_DEBUG("Found %d child joints:", joints.size());
+  ROS_DEBUG("Found %lu child joints:", joints.size());
   for (joint_list::const_iterator it=joints.begin(); it!=joints.end(); ++it)
   {
     ROS_DEBUG_STREAM("  " << (*it)->name << ": type " <<  (*it)->type);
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
A bit pendantic maybe, but the assumption in the proposed change is that we'll never have more than 2^32 links and joints. Perhaps it would be more correct to change the format specifier?

`size_t` is `long unsigned int`: `%lu`?

VictorLamoine:
`long unsigned int` it is! You are right let's stick to the type.

industrial_utils/src/utils.cpp
```diff
@@ -78,7 +78,7 @@ bool findChainJointNames(const boost::shared_ptr<const urdf::Link> &link, bool i
 
   // check for joints directly connected to this link
   const joint_list &joints = link->child_joints;
-  ROS_DEBUG("Found %d child joints:", joints.size());
+  ROS_DEBUG("Found %lu child joints:", (long unsigned int) joints.size());
```
gavanderhoorn: Is the typecast still necessary with the `%lu` format specifier?

VictorLamoine: Not annymore

shaun-edwards:
Looks good to me... @gavanderhoorn, +1?

gavanderhoorn:
I'm not sure, but isn't `sizeof(size_t) == sizeof(long unsigned int)` architecture specific? That might be one thing to verify. If that is the case, explicit casting (as @VictorLamoine had done earlier), might actually be a safer fix.

Otherwise, +1.

Issue #85, opened 2014-10-22
joint_traj_action: segfault (null pointer) in withinGoalConstraints() when no feedback has been received


gavanderhoorn:
While looking into [Rename JOINT_NAMES in ur_driver test_move.py](http://answers.ros.org/question/195376/rename-joint_names-in-ur_driver-test_movepy/) on ROS Answers:

I've observed that `joint_trajectory_action::withinGoalConstraints(..)` tries to dereference the `last_trajectory_state_` member variable in a call to `industrial_robot_client::utils::isWithinRange(..)`, even when no `FollowJointTrajectoryFeedback` has ever been received.

Sending a goal to the `joint_trajectory_action` node "too early" (ie: `JointTrajectoryAction::controllerStateCB(..)` has not been invoked) then results in a segfault.

Backtrace:

``` shell
[1413970334.414501639] [ INFO] [/joint_trajectory_action]: Received new goal
joint_trajectory_action: /usr/include/boost/smart_ptr/shared_ptr.hpp:418: T* boost::shared_ptr<T>::operator->() const [with T = const control_msgs::FollowJointTrajectoryFeedback_<std::allocator<void> >]: Assertion `px != 0' failed.

Program received signal SIGABRT, Aborted.
0xb7fdf1b2 in ?? () from /lib/ld-linux.so.2
(gdb) bt
#0  0xb7fdf1b2 in ?? () from /lib/ld-linux.so.2
#1  0xb7a81e0f in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xb7a85455 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0xb7a7acb5 in ?? () from /lib/i386-linux-gnu/libc.so.6
#4  0xb7a7ad67 in __assert_fail () from /lib/i386-linux-gnu/libc.so.6
#5  0x080c7a6d in boost::shared_ptr<control_msgs::FollowJointTrajectoryFeedback_<std::allocator<void> > const>::operator-> (this=0xbfffe838)
    at /usr/include/boost/smart_ptr/shared_ptr.hpp:418
#6  0x080bf571 in industrial_robot_client::joint_trajectory_action::JointTrajectoryAction::withinGoalConstraints (this=0xbfffe650, msg=..., traj=...)
    at /home/user/ros/answers_test_hydro/src/industrial_core/industrial_robot_client/src/joint_trajectory_action.cpp:293
[...]

PR #86, opened 2014-10-23
Fix for issue #85: reject all goals until feedback has been received

gavanderhoorn:
As per subject.


gavanderhoorn:
Ping? Bit impatient, I know, but I just ran into this myself as well.

shaun-edwards:
+1

PR #87, opened 2014-10-30
Added support for gethostbyname, for passing host names in addition to I...

frederickproctor:
...P addresses.

### DIFF
--- a/simple_message/include/simple_message/socket/tcp_socket.h
+++ b/simple_message/include/simple_message/socket/tcp_socket.h
@@ -42,6 +42,7 @@
 
 #ifdef LINUXSOCKETS
 #include "sys/socket.h"
+#include "netdb.h"		/* gethostbyname */
 #include "arpa/inet.h"
 #include "string.h"
 #include "unistd.h"
--- a/simple_message/include/simple_message/socket/tcp_socket.h
+++ b/simple_message/include/simple_message/socket/tcp_socket.h
@@ -42,7 +42,7 @@
 
 #ifdef LINUXSOCKETS
 #include "sys/socket.h"
-#include "netdb.h"		/* gethostbyname */
+#include "netdb.h"
 #include "arpa/inet.h"
 #include "string.h"
 #include "unistd.h"
--- a/simple_message/src/socket/tcp_client.cpp
+++ b/simple_message/src/socket/tcp_client.cpp
@@ -57,6 +57,8 @@ bool TcpClient::init(char *buff, int port_num)
   int rc;
   bool rtn;
   int disableNodeDelay = 1;
+  struct hostent *ent;
+  struct in_addr *in_a;
 
   rc = SOCKET(AF_INET, SOCK_STREAM, 0);
   if (this->SOCKET_FAIL != rc)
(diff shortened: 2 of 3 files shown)


shaun-edwards:
Thanks for the contribution!

One small modification is requested.  For some history, this simple message library tries to be OS independent (at one time it ran on the Motoman controller, and in the future it would be nice to use it on Windows).

All system calls are wrapped in macros (for linux see [here](https://github.com/frederickproctor/industrial_core/blob/gethostbyname/simple_message/include/simple_message/socket/simple_socket.h#L54-L71), for motoplus see [here](https://github.com/frederickproctor/industrial_core/blob/gethostbyname/simple_message/include/simple_message/socket/simple_socket.h#L81-L112)) to achieve this OS independence.  

Could you please wrap the call to `gethostbyname` in a macro.  The linux version would perform the call the way you have it.  The motoplus version would simply return `NULL` since I don't believe that it is supported.

Other style comments in line.

simple_message/include/simple_message/socket/tcp_socket.h
```diff
@@ -42,6 +42,7 @@
 
 #ifdef LINUXSOCKETS
 #include "sys/socket.h"
+#include "netdb.h"		/* gethostbyname */
```
shaun-edwards: No need to have the comment `/* gethostbyname */`or to at least match the current style.

simple_message/src/socket/tcp_client.cpp
```diff
@@ -70,10 +72,16 @@ bool TcpClient::init(char *buff, int port_num)
       LOG_WARN("Failed to set no socket delay, sending data can be delayed by up to 250ms");
     }
 
+
     // Initialize address data structure
     memset(&this->sockaddr_, 0, sizeof(this->sockaddr_));
     this->sockaddr_.sin_family = AF_INET;
-    this->sockaddr_.sin_addr.s_addr = INET_ADDR(buff);
+    if (NULL != (ent = gethostbyname(buff))) {
```
shaun-edwards: Looks like the convention in this file is to put the `{` on the next line.

simple_message/src/socket/tcp_client.cpp
```diff
@@ -70,10 +72,16 @@ bool TcpClient::init(char *buff, int port_num)
...
 
+
     // Initialize address data structure
     memset(&this->sockaddr_, 0, sizeof(this->sockaddr_));
     this->sockaddr_.sin_family = AF_INET;
-    this->sockaddr_.sin_addr.s_addr = INET_ADDR(buff);
+    if (NULL != (ent = gethostbyname(buff))) {
+      in_a = (struct in_addr *) ent->h_addr_list[0];
+      this->sockaddr_.sin_addr.s_addr = in_a->s_addr;
+    } else {
```
shaun-edwards: Same here for `{`

frederickproctor: Thanks for your coding comments, Shaun. I made the edits you noted, to wit: 
1. Macro'ed out gethostbyname in both the Linux and Motoman sections of simple_socket.h, with the Motoman resulting in NULL. Called that macro in tcp_client.c.
2. Took out the /\* gethostbyname*/ comment in the header, for consistency. 
3. Fixed the brace style in tcp_client.c, to be a new line for braces. I updated my emacs indenting accordingly. 
I think the first pull request is still pending, and you should see this new commit. 
--Fred

shaun-edwards:
Thanks for the contribution @frederickproctor.  +1.

Issue #63, opened 2014-01-07
robot_simulator: redundant roslib.load_manifest(..)


gavanderhoorn:
As far as I understand [catkin/what#Python Modules (Libraries)](http://wiki.ros.org/catkin/what#Python_Modules_.28Libraries.29) on the ROS wiki, Python scripts no longer need to load the manifest. Ie:

`roslib.load_manifest('industrial_robot_simulator')`

Is redundant.

PR #91, opened 2014-10-31
Fix for issue #63: remove redundant roslib.load_manifest() from robot_simulator

gavanderhoorn:
As per subject.

No longer necessary according to [catkin/what#Python Modules (Libraries)](http://wiki.ros.org/catkin/what#Python_Modules_.28Libraries.29).


gavanderhoorn:
Ping?

shaun-edwards:
+1...Thanks for the contribution.

PR #92, opened 2014-11-15
robot_simulator: quiet down node (use logdebug()).

gavanderhoorn:
As per subject.

Node was rather verbose by default. Now requires increasing log level to debug.


shaun-edwards:
+1

PR #93, opened 2014-11-15
Add GetRobotInfo service server implementation to Robot Simulator

gavanderhoorn:
Users can supply the information that should be returned by the server by configuring the private parameter `~robot_info` in a launch file like so (the information could also be loaded from a yaml file, again using `rosparam`):

``` xml
  <node .. type="industrial_robot_simulator" name=".." ..>
    <rosparam param="robot_info">
      controller:
        model: 'My Robot Controller'
        serial_number: 'ABC-123'
        sw_version: 'V1.23/ABC'
      robots:
        - model: 'My Robot Model'
          serial_number: '123456'
          sw_version: 'V4.56/DEF'
    </rosparam>
  </node>
```

As the Industrial Robot Simulator currently simulates only a single robot / motion group, specifying multiple robot entries will result in only the first of the list to be used (and a warning being emitted).

Fields absent from the `rosparam` parameter will result in empty strings being assigned to the corresponding fields in the `controller` and `robot` parts of the service response.

### DIFF
--- a/industrial_robot_simulator/industrial_robot_simulator
+++ b/industrial_robot_simulator/industrial_robot_simulator
@@ -14,8 +14,11 @@ from industrial_msgs.msg import RobotStatus
 from trajectory_msgs.msg import JointTrajectory
 from trajectory_msgs.msg import JointTrajectoryPoint
 
+# Services
+from industrial_msgs.srv import GetRobotInfo, GetRobotInfoResponse
+
 # Reference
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
I squashed the whole thing into one commit. Bit cleaner like this.

shaun-edwards:
+1...Thanks for the contribution!

gavanderhoorn:
Np. I think this is actually the first 'robot' to support this service :).

PR #98, opened 2014-12-16
Fill stamp of the RobotStatus message Fix: #97

liborw:
Just edited on a github didn't test it.


gavanderhoorn:
+1 from me.

Ideally this stamp would also be a field in the Simple Message message structure, so the controller can provide it, but that is not currently possible.

Issue #99, opened 2015-01-15
robot_simulator: explicitly set queue_size for publishers


gavanderhoorn:
Starting the `industrial_robot_simulator` node in Indigo results in the following warnings being printed:

```
industrial_robot_simulator/industrial_robot_simulator:221: SyntaxWarning: The publisher should be created with an explicit keyword argument 'queue_size'. Please see http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers for more information.
  self.joint_state_pub = rospy.Publisher('joint_states', JointState)
industrial_robot_simulator/industrial_robot_simulator:225: SyntaxWarning: The publisher should be created with an explicit keyword argument 'queue_size'. Please see http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers for more information.
  self.joint_feedback_pub = rospy.Publisher('feedback_states', FollowJointTrajectoryFeedback)
industrial_robot_simulator/industrial_robot_simulator:228: SyntaxWarning: The publisher should be created with an explicit keyword argument 'queue_size'. Please see http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers for more information.
  self.robot_status_pub = rospy.Publisher('robot_status', RobotStatus)
```

Related links:
- [[ros-users] RFC: warning for (lacking) queue_size for python publishers in Indigo?](http://lists.ros.org/pipermail/ros-users/2014-March/068333.html)
- [rospy/Overview/Publishers and Subscribers - 1.3 queue_size: publish() behavior and queuing](http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers#queue_size:_publish.28.29_behavior_and_queuing)

PR #100, opened 2015-01-15
FIx for issue #99: explicit queue size for Publishers in robot simulator

gavanderhoorn:
As per subject.

This is backwards compatible with Hydro (and Groovy).


shaun-edwards:
+1.

Issue #48, opened 2013-12-27
simple_message: caller should pass errno as parameter to logSocketError(..)


gavanderhoorn:
The implementation of `logSocketError(..)` in `simple_socket.h` is naive: the value of `errno` is retrieved inside the function (see [simple_socket.h#L210](https://github.com/ros-industrial/industrial_core/blob/groovy-devel/simple_message/include/simple_message/socket/simple_socket.h#L210)). Depending on the context in which this function is called, `errno` could've already changed, making the printed message useless.

A better approach would be to pass `errno` as a parameter to the `logSocketError(..)` function, making it the responsibility of the caller to ensure the validity of the `errno` value.

shaun-edwards:
Fixed in #102.

PR #102, opened 2015-02-06
Fix for issue #48.  Deprecated original logSocketError method

shaun-edwards:


shaun-edwards:
@gavanderhoorn - please review.

gavanderhoorn:
I think you missed `tcp_client.cpp` and `udp_socket.cpp`.

Also, according to [gcc/onlinedocs/C Extensions - 6.31 - Declaring Attributes of Functions](https://gcc.gnu.org/onlinedocs/gcc/Function-Attributes.html), `deprecated` takes a string argument, which we should perhaps use to point users to the replacement function?

shaun-edwards:
Added deprecated message and squashed commit.  Method is not used in `udp_socket.cpp` (odd, but OK).

gavanderhoorn:
`udp_socket.cpp` has a single call to `logSocketError(..)` in `bool UdpSocket::rawPoll(int timeout, bool & ready, bool & error)` on [line 155](https://github.com/ros-industrial/industrial_core/blob/e00a9f7a9d346450c7255b339fc4863fd59ba5b4/simple_message/src/socket/udp_socket.cpp#L155).

shaun-edwards:
OK...so the problem was that the `indigo-devel` branch was not synced with `hydro-devel`.  This has now been fixed.  All `logSocketError` methods have been updated

shaun-edwards:
@gavanderhoorn, please review (hopefully for the last time).

gavanderhoorn:
lgtm, +1.

shaun-edwards:
@gavanderhoorn, thanks for the review (and your patience).

PR #105, opened 2015-02-28
Issue #2 - correctly initialized connected state for udp connections

shaun-edwards:
Also, this fixes the udp unit tests.  I'm not sure why they were broken.

### DIFF
--- a/simple_message/include/simple_message/socket/simple_socket.h
+++ b/simple_message/include/simple_message/socket/simple_socket.h
@@ -144,7 +144,12 @@ public:
   /**
      * \brief Constructor
      */
-  SimpleSocket(){}
+  SimpleSocket()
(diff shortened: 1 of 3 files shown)


shaun-edwards:
@jrgnicho, can you review ( I know it's easy ).

shaun-edwards:
@jrgnicho, ping?

jrgnicho:
The only change here sets an inherited bool member to false, should the right fix be to initialize that member to false in the constructor of the base class?

shaun-edwards:
@jrgnicho, all common constructor code moved to base class.

shaun-edwards:
@jrgnicho, ping.

jrgnicho:
+1 for merge

Issue #89, opened 2014-10-31
robot_simulator: strange package dependencies


gavanderhoorn:
The `CMakeLists.txt` and package manifest of the `industrial_robot_simulator` package contains some (afaict) strange dependency declarations. Seeing as this is a Python only node (with only a ROS API level interface), I'm wondering why the `roscpp` and `industrial_robot_client` packages are listed as catkin / build & run deps.

gavanderhoorn:
This was fixed when #106 was merged.

Issue #45, opened 2013-12-04
CMake targets should be prefixed by ${PROJECT_NAME}


gavanderhoorn:
See ros-industrial/ros_industrial_issues#8.

gavanderhoorn:
This is actually more involved than just prefixing all targets with the project_name variable due to the way `simple_message` and `industrial_robot_client` build and export their libraries (#46). The fact that CMake resolves linking dependencies at configuration time makes it impossible to depend on renamed targets (ie, depend on `foo` instead of `foo-before-rename`).

Without exporting the `simple_message` and `industrial_robot_client` library targets in the `catkin_package(..)` statement, end-users will have to explicitly depend on `industrial_robot_client_industrial_robot_client` or `simple_message_simple_message_bswap`. It is probably best to first sort out the issues described in #46.

shaun-edwards:
This was scheduled for `indigo`, but it only applies to `simple_message` and `industrial_robot_client` which cannot be fixed as stated above.

gavanderhoorn:
This is not an issue for build tools like `catkin_make_isolated`, `catkin_tools` or Colcon.

The ROS buildfarm also doesn't run into any issues with this as it builds everything in isolation.

Prefixing the targets would only help `catkin_make`, but even there recent CMake versions isolate targets to be defined per directory/file.

As such: closing.

PR #106, opened 2015-02-28
Issue #89 - fixes industrial_robot_simulator depends

shaun-edwards:
Depends on PR #105

### DIFF
--- a/industrial_robot_simulator/CMakeLists.txt
+++ b/industrial_robot_simulator/CMakeLists.txt
@@ -2,22 +2,23 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(industrial_robot_simulator)
 
-find_package(catkin REQUIRED COMPONENTS roscpp std_msgs sensor_msgs 
-  control_msgs trajectory_msgs industrial_robot_client)
+find_package(catkin REQUIRED COMPONENTS std_msgs sensor_msgs
+  control_msgs trajectory_msgs)
 
 catkin_package(
-    CATKIN_DEPENDS roscpp std_msgs sensor_msgs control_msgs trajectory_msgs industrial_robot_client
+    CATKIN_DEPENDS std_msgs sensor_msgs control_msgs trajectory_msgs
 )
 
 ##########
 ## Test ##
 ##########
 
-# ROS launch testing
-## ROS launch testing should be re-enabled when the roslaunch script is officially
-## released (see https://github.com/ros/ros_comm/pull/333)
-##find_package(roslaunch)
-##roslaunch_add_file_check(launch)
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
+  roslaunch_add_file_check(launch
+    DEPENDENCIES
+      joint_trajectory_action)
+endif()
 
 #############
 ## Install ##
--- a/industrial_robot_simulator/CMakeLists.txt
+++ b/industrial_robot_simulator/CMakeLists.txt
@@ -17,7 +17,7 @@ if (CATKIN_ENABLE_TESTING)
   find_package(roslaunch REQUIRED)
   roslaunch_add_file_check(launch
     DEPENDENCIES
-      industrial_core)
+      joint_trajectory_action)
 endif()
 
 #############
--- a/industrial_robot_simulator/CMakeLists.txt
+++ b/industrial_robot_simulator/CMakeLists.txt
@@ -17,7 +17,7 @@ if (CATKIN_ENABLE_TESTING)
   find_package(roslaunch REQUIRED)
   roslaunch_add_file_check(launch
     DEPENDENCIES
-      joint_trajectory_action)
+      industrial_core)
 endif()
 
 #############
--- a/industrial_robot_simulator/package.xml
+++ b/industrial_robot_simulator/package.xml
@@ -22,4 +22,5 @@
   <run_depend>industrial_robot_client</run_depend>
   <run_depend>industrial_msgs</run_depend>
   <run_depend>python-rospkg</run_depend>
+  <run_depend>rospy</run_depend>
 </package>
--- a/industrial_robot_simulator/package.xml
+++ b/industrial_robot_simulator/package.xml
@@ -14,7 +14,7 @@
   <build_depend>control_msgs</build_depend>
   <build_depend>trajectory_msgs</build_depend>
   <build_depend>industrial_msgs</build_depend>
-  <build_depend>python-rospkg</build_depend>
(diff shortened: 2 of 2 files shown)


shaun-edwards:
@gavanderhoorn, could you review.  This addresses your issue #89.  Please verify the dependencies are now correct.  I wasn't quite sure if python dependencies should be `run_depend` only.

shaun-edwards:
@gavanderhoorn, ping/help.

gavanderhoorn:
I don't believe we need all the `build_depend`s in there (at least, not according to [Python module dependencies](http://docs.ros.org/indigo/api/catkin/html/howto/format1/python_module_dependencies.html) from the catkin documentation).

Must say it's becoming increasingly unclear to me what the correct approach is for Python pkgs, but I believe only `run_depend`s are required, unless you are depending on a message pkg.

industrial_robot_simulator/package.xml
```diff
@@ -9,15 +9,12 @@
...
   <build_depend>sensor_msgs</build_depend>
   <build_depend>control_msgs</build_depend>
   <build_depend>trajectory_msgs</build_depend>
-  <build_depend>industrial_robot_client</build_depend>
   <build_depend>industrial_msgs</build_depend>
   <build_depend>python-rospkg</build_depend>
-  <run_depend>roscpp</run_depend>
   <run_depend>std_msgs</run_depend>
   <run_depend>sensor_msgs</run_depend>
   <run_depend>control_msgs</run_depend>
```
jrgnicho, line 20: Shoulnd't rospy be at least listed as a run_depend?

gavanderhoorn, line 20: Yes, that should definitely be there.

industrial_robot_simulator/CMakeLists.txt
```diff
@@ -2,22 +2,23 @@ cmake_minimum_required(VERSION 2.8.3)
...
-# ROS launch testing
-## ROS launch testing should be re-enabled when the roslaunch script is officially
-## released (see https://github.com/ros/ros_comm/pull/333)
-##find_package(roslaunch)
-##roslaunch_add_file_check(launch)
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
+  roslaunch_add_file_check(launch
+    DEPENDENCIES
+      joint_trajectory_action)
```
gavanderhoorn, line 20: This would be the `industrial_core` one then?

industrial_robot_simulator/package.xml
shaun-edwards, line 20: Added (see below)

shaun-edwards:
Still depends on PR #105.  I think some of the `build_depends` are captured as inherited dependencies and thus not needed.  For now I'd like to leave them in.

industrial_robot_simulator/CMakeLists.txt
```diff
@@ -2,22 +2,23 @@ cmake_minimum_required(VERSION 2.8.3)
...
 ## Test ##
 ##########
 
-# ROS launch testing
-## ROS launch testing should be re-enabled when the roslaunch script is officially
-## released (see https://github.com/ros/ros_comm/pull/333)
-##find_package(roslaunch)
-##roslaunch_add_file_check(launch)
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
```
gavanderhoorn, line 17: Adding this will require a `build_depend` on `roslaunch` in the manifest. I suggest:

``` xml
<build_depend version_gte="1.9.55">roslaunch</build_depend>
```

As you're also using the `DEPENDENCIES` arg, which is only supported from that version onwards.

industrial_robot_simulator/CMakeLists.txt
```diff
@@ -2,22 +2,23 @@ cmake_minimum_required(VERSION 2.8.3)
...
+  roslaunch_add_file_check(launch
+    DEPENDENCIES
+      industrial_core)
```
gavanderhoorn: `DEPENDENCIES` only takes a list of target names, not package names.

industrial_robot_simulator/package.xml
```diff
@@ -9,15 +9,12 @@
...
 
   <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
   <build_depend>std_msgs</build_depend>
   <build_depend>sensor_msgs</build_depend>
   <build_depend>control_msgs</build_depend>
   <build_depend>trajectory_msgs</build_depend>
-  <build_depend>industrial_robot_client</build_depend>
   <build_depend>industrial_msgs</build_depend>
   <build_depend>python-rospkg</build_depend>
```
gavanderhoorn: `python-rospkg` can be deleted. Only `run_depend` needed (see [Python module dependencies](http://docs.ros.org/indigo/api/catkin/html/howto/format1/python_module_dependencies.html)).

industrial_robot_simulator/CMakeLists.txt
shaun-edwards: `joint_trajectory_action` was the right target then...see [here](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/industrial_robot_client/CMakeLists.txt#L134).  It should be prefixed by the package name, but this is really another issue.

gavanderhoorn: Yep, that seems to be the right one.

And, re: prefixing targets: yes, see #45.

gavanderhoorn: Although btw, it should be possible to depend on `${industrial_core_EXPORTED_TARGETS}`, iirc.

shaun-edwards:
@gavanderhoorn, can I merge as is?

gavanderhoorn:
No; see in-line comments (args to `DEPENDENCIES` and `python-rospkg` build depend).

shaun-edwards:
OK...failed to push last time...all changes should be in this PR now.

industrial_robot_simulator/package.xml
```diff
@@ -25,4 +22,5 @@
   <run_depend>industrial_robot_client</run_depend>
```
gavanderhoorn, line 22: One last question: why is this needed? That is a C++ only pkg, is it not?

shaun-edwards, line 22: There is a launch dependency...[here](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/industrial_robot_simulator/launch/robot_interface_simulator.launch)

gavanderhoorn, line 22: Yep, forgot about that.

gavanderhoorn:
+1.

PR #113, opened 2015-06-03
Fix rejected actions when goal requests arrive inbetween watchdog reset und arrival of next status package

### LINKED ISSUE #112: Industrial_robot_client: Greedy Watchdog leads to rejection of valid goal
Hi,

this is a regression due to #86 / #85 . Sometimes execution of valid trajectories is rejected with this message:
Joint trajectory action rejected: waiting for (initial) feedback from controller

I poll my robot with 20 Hz (arbitrary), the watchdog is set to 1Hz. So I think the following happens:

1230s: JointTrajectoryAction::controllerStateCB: trajectory_state_recvd_ = true;
1230.01s: JointTrajectoryAction::watchdog: trajectory_state_recvd_ = false;
1230.02s: JointTrajectoryAction::goalCB -> rejected
(1230.03s: JointTrajectoryAction::goalCB -> rejected)
(1230.04s: JointTrajectoryAction::goalCB -> rejected)
1230.05s: JointTrajectoryAction::controllerStateCB: trajectory_state_recvd_ = true; but too late!

(the events in brackets are due to me retrying the execution if it doesn't work)

So I think the watchdog should manage some other variable like "controller_alive_" which is then checked by goalCB. trajectory_state_recvd_ is meant as an internal variable to the watchdog.

As a workaround I can of course increase/add a sleep between successive attempts to send the goal.

simonschmeisser:
this fixes #112  , a regression due to #86 / #85 . Sometimes execution of valid trajectories is rejected with this message:
Joint trajectory action rejected: waiting for (initial) feedback from controller

I poll my robot with 20 Hz (arbitrary), the watchdog is set to 1Hz. So I think the following happens:

0s: JointTrajectoryAction::controllerStateCB: trajectory_state_recvd_ = true;
0.01s: JointTrajectoryAction::watchdog: trajectory_state_recvd_ = false;
0.02s: JointTrajectoryAction::goalCB -> rejected
(0.03s: JointTrajectoryAction::goalCB -> rejected)
(0.04s: JointTrajectoryAction::goalCB -> rejected)
0.05s: JointTrajectoryAction::controllerStateCB: trajectory_state_recvd_ = true; but too late!

(the events in brackets are due to me retrying the execution if it doesn't work)

So I think the watchdog should manage some other variable like "controller_alive_" which is then checked by goalCB. trajectory_state_recvd_ is meant as an internal variable to the watchdog.

### DIFF
--- a/industrial_robot_client/src/joint_trajectory_action.cpp
+++ b/industrial_robot_client/src/joint_trajectory_action.cpp
@@ -39,12 +39,13 @@ namespace industrial_robot_client
 namespace joint_trajectory_action
 {
 
-const double JointTrajectoryAction::WATCHD0G_PERIOD_ = 1.0;
+const double JointTrajectoryAction::WATCHDOG_PERIOD_ = 1.0;
 const double JointTrajectoryAction::DEFAULT_GOAL_THRESHOLD_ = 0.01;
 
 JointTrajectoryAction::JointTrajectoryAction() :
     action_server_(node_, "joint_trajectory_action", boost::bind(&JointTrajectoryAction::goalCB, this, _1),
-                   boost::bind(&JointTrajectoryAction::cancelCB, this, _1), false), has_active_goal_(false)
+                   boost::bind(&JointTrajectoryAction::cancelCB, this, _1), false), has_active_goal_(false),
+                       controller_alive_(false)
 {
   ros::NodeHandle pn("~");
 
@@ -62,7 +63,7 @@ JointTrajectoryAction::JointTrajectoryAction() :
   sub_trajectory_state_ = node_.subscribe("feedback_states", 1, &JointTrajectoryAction::controllerStateCB, this);
   sub_robot_status_ = node_.subscribe("robot_status", 1, &JointTrajectoryAction::robotStatusCB, this);
 
-  watchdog_timer_ = node_.createTimer(ros::Duration(WATCHD0G_PERIOD_), &JointTrajectoryAction::watchdog, this);
+  watchdog_timer_ = node_.createTimer(ros::Duration(WATCHDOG_PERIOD_), &JointTrajectoryAction::watchdog, this, true);
   action_server_.start();
 }
 
@@ -82,34 +83,27 @@ void JointTrajectoryAction::watchdog(const ros::TimerEvent &e)
   {
     ROS_DEBUG("Waiting for subscription to joint trajectory state");
   }
-  if (!trajectory_state_recvd_)
-  {
-    ROS_DEBUG("Trajectory state not received since last watchdog");
-  }
+
+  ROS_WARN("Trajectory state not received for %f seconds", WATCHDOG_PERIOD_);
+  controller_alive_ = false;
+
 
   // Aborts the active goal if the controller does not appear to be active.
   if (has_active_goal_)
   {
-    if (!trajectory_state_recvd_)
+    // last_trajectory_state_ is null if the subscriber never makes a connection
+    if (!last_trajectory_state_)
     {
-      // last_trajectory_state_ is null if the subscriber never makes a connection
-      if (!last_trajectory_state_)
-      {
-        ROS_WARN("Aborting goal because we have never heard a controller state message.");
-      }
-      else
-      {
-        ROS_WARN_STREAM(
-            "Aborting goal because we haven't heard from the controller in " << WATCHD0G_PERIOD_ << " seconds");
-      }
-
-      abortGoal();
-
+      ROS_WARN("Aborting goal because we have never heard a controller state message.");
     }
-  }
+    else
+    {
+      ROS_WARN_STREAM(
+          "Aborting goal because we haven't heard from the controller in " << WATCHDOG_PERIOD_ << " seconds");
+    }
 
-  // Reset the trajectory state received flag
-  trajectory_state_recvd_ = false;
+    abortGoal();
+  }
 }
 
 void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle & gh)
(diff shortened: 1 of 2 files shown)


industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -216,9 +210,12 @@ void JointTrajectoryAction::cancelCB(JointTractoryActionServer::GoalHandle & gh)
...
 void JointTrajectoryAction::controllerStateCB(const control_msgs::FollowJointTrajectoryFeedbackConstPtr &msg)
 {
-  //ROS_DEBUG("Checking controller state feedback");
+  ROS_DEBUG("Checking controller state feedback");
   last_trajectory_state_ = msg;
-  trajectory_state_recvd_ = true;
+  controller_alive_ = true;
+
+  watchdog_timer_.stop();
+  watchdog_timer_.start();
```
gavanderhoorn, line 218: afaict, it should be possible to 'restart' a timer using `setPeriod(..)` (eventually in [TimerManager](http://docs.ros.org/indigo/api/roscpp/html/timer__manager_8h_source.html#l00420)). That would remove the need for the `stop()`, `start()` pair.

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -62,7 +62,8 @@ JointTrajectoryAction::JointTrajectoryAction() :
   sub_trajectory_state_ = node_.subscribe("feedback_states", 1, &JointTrajectoryAction::controllerStateCB, this);
   sub_robot_status_ = node_.subscribe("robot_status", 1, &JointTrajectoryAction::robotStatusCB, this);
 
-  watchdog_timer_ = node_.createTimer(ros::Duration(WATCHD0G_PERIOD_), &JointTrajectoryAction::watchdog, this);
+  controller_alive_ = false;
```
gavanderhoorn: Variable should be initialised in initialisation list.

industrial_robot_client/src/joint_trajectory_action.cpp
gavanderhoorn, line 218: Not sure about one-shot timers though.

simonschmeisser, line 218: that feels like using a side-effect, stop start is a bit clumsy but quite legible, but I have no problem with being overvoted

gavanderhoorn, line 218: You're right, it is a side-effect. Explicit start/stop is better.

industrial_robot_client/src/joint_trajectory_action.cpp
simonschmeisser: fixed

shaun-edwards:
+1.... @simonschmeisser, thanks for the contribution.

gavanderhoorn:
@shaun-edwards: you merged this PR into the `indigo` release branch, not into `indigo-devel` (the PR was setup to do that, so not your fault).

simonschmeisser:
I was just using the branch with the most recent commits, maybe I misunderstood the branching concept used. Shouldn't there in theory be a jade-devel branch since indigo was released already? Is this documented somewhere?

gavanderhoorn:
> Shouldn't there in theory be a jade-devel branch since indigo was released already?

Are you implying that we don't develop for a ROS version anymore after it's been released :)?

The `-devel` suffix is used in many ROS related repositories, and is used to indicate that all development is targeted at (or compatible with) a specific ROS release. It does _not_ mean "developments to get an X release ready".

'Best practice' suggests that a repository should only branch for a specific ROS release if there is a need for it. Introducing changes incompatible with the previous release would be such a need. There is nothing in `industrial_core` right now that isn't Jade compatible, so it would follow that there is no real need for a branch. Branching would only increase the maintenance burden, as we'd need to keep all branches in sync.

Releases are then created by tagging the `-devel` branch. In ROS-I specifically, we adopted so called _release branches_, which track the development branches but do not mirror them (they include all changes up-to a tagged release).

As to documentation: it is a bit hidden, but the _Install_ page does refer to it:

> The ROS-Industrial git repository is organized similar to the ROS repository conventions, with one exception. ROS-Industrial releases (stable) are made from _release name_ branches. _release name_-devel branches are reserved for continued development (potentially unstable). :
> 
> ```
> repository
>   ...tags                     (all released versions)
>   ...branches
>     ...<release_name>         (latest release - matches documentation)
>     ...<release_name>-devel   (latest development - may not match documentation)
> ```

from [wiki/Industrial/Install - ROS-Industrial Stacks - Source Installation (Recommended for Developers) - Source Repository Layout](http://wiki.ros.org/Industrial/Install#Industrial.2BAC8-Install.2BAC8-Hydro.Source_Repository_Layout).

gavanderhoorn:
O and just to make sure: this is also not 'your fault'. We should've just checked more carefully how the PR was created.

simonschmeisser:
ok, thanks, now I get the concept. Don't worry, I didn't feel offended. Do I need to change anything or will you be able to fix things?

Just some more questions while we're at it: Binary releases (dpkg) happen from <release_name> branch? Do fixes get backported typically?

shaun-edwards:
This is definitely my fault. I should have checked (I thought I did but apparently not).

gavanderhoorn:
@shaun-edwards: perhaps @simonschmeisser's question is a good reason to create some _Branching Policy_ documentation and / or make a separate page out of that section on the _Install_ page about source repository layout.

gavanderhoorn:
Also: @simonschmeisser: this is easily fixed, and you don't need to do anything.

shaun-edwards:
I'll fix this when I get into the office. Does everyone think the wiki is the right place. The "GitHub" way is to have a CONTRIBUTING readme.

gavanderhoorn:
Hm, using the contribution hook is a nice idea. I've only seen it used to draw attention to any licensing requirements, but I guess you could put this kind of info into it as well. I created an issue for that some time ago over at ros-industrial/ros_industrial_issues#26.

I'd still make a page on the wiki though: that way it is a stand-alone document we can refer ppl to, not a file in a vcs somewhere. It would be similar to the PR Review Guidelines.

shaun-edwards:
OK...that was ugly (and possibly way more complicated than it needed to be).  I have successfully reverted the merge into indigo.  In order to avoid merge problems down the road, I had to sync indigo and indigo-devel.  I then reverted the revert in indigo-devel.  The correct code is in indigo-devel and I have verified a PR from indigo-devel into indigo catches these changes.  In case you are curious, I was trying to avoid [this](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt).

PR #115, opened 2015-06-24
Issue#114 - Action Server withinGoalConstraints check removal

Jmeyer1292:
This PR removes the goal constraints check from the goal-acceptance function of the industrial action server. Because of a previous check to size of the trajectory, this effectively removes only the check on whether the first and last points are the same. 

The robot continues to check for goal completion in the callback, but that should behave correctly. In my own testing, this pull request has fixed the issues I was experiencing.

### DIFF
--- a/industrial_robot_client/src/joint_trajectory_action.cpp
+++ b/industrial_robot_client/src/joint_trajectory_action.cpp
@@ -139,28 +139,16 @@ void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle & gh)
         ROS_WARN("Received new goal, canceling current goal");
         abortGoal();
       }
+      
+      gh.setAccepted();
+      active_goal_ = gh;
+      has_active_goal_ = true;
 
-      // Sends the trajectory along to the controller
-      if (withinGoalConstraints(last_trajectory_state_, gh.getGoal()->trajectory))
-      {
+      ROS_INFO("Publishing trajectory");
 
-        ROS_INFO_STREAM("Already within goal constraints, setting goal succeeded");
-        gh.setAccepted();
-        gh.setSucceeded();
-        has_active_goal_ = false;
-
-      }
-      else
-      {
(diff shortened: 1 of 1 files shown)


shaun-edwards:
@JeremyZoss, @gavanderhoorn, any thoughts?

I suggested this change to @Jmeyer1292.  The check is a hold over from the original PR2 code, which was pick and place focused.  As we move to more process based paths, this issue could occur more often.

gavanderhoorn:
Change seems to make sense. I'd say that any action server that performs a similar check ("is last point of trajectory within goal constraints?") will encounter the same issue with trajectories that end where they start.

Should we take a look at `JointTrajectoryAction::controllerStateCB` too? Seems to follow the same logic ([here](https://github.com/ros-industrial/industrial_core/blob/49ff2d214e6ee471e104eeb86e54730fb448d6dd/industrial_robot_client/src/joint_trajectory_action.cpp#L244-L277))? Or do we rely on `RobotStatus.msg::in_motion` to catch that?

Should we backport this to Hydro as well? Seems like a rather critical bug to me. The fix would be behaviour changing though.

JeremyZoss:
+1.  This just pushes the burden for "check if we're already at goal" back on the application code, which is probably a better fit anyway.

@gavanderhoorn : I think the `controllerStateCB` logic needs to stay as-is, for now.  If the robot driver doesn't support `in_motion`, this is our only way to detect end-of-trajectory.  We could add something to `ind_rbt_client` to support trajectory status-feedback, but this would be difficult for trajectories that are downloaded all-at-once to the robot.  It'd work fine for streaming trajectories, as the `ind_rbt_client` (mostly) knows when the trajectory is done.

shaun-edwards:
For now I'm going to push this into Indigo only.  If this change causes an issue and we push it into Hydro, I'm afraid we would miss the last sync and end up with a permanently "broken" hydro debian.

Issue #114, opened 2015-06-15
Premature termination of trajectory action


Jmeyer1292:
Hello all,

The ros-industrial client's action server interface performs checks for 'goal completion' upon accepting a new trajectory and as it gets feedback about the robot state. See [this link](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/industrial_robot_client/src/joint_trajectory_action.cpp#L244). 

Unfortunately this means that any trajectory starting at the 'end point' is immediately considered successfully executed. Furthermore any trajectory that passes through the end point, even if there are more points to go, will be considered successfully executed.

If I'm using the system wrong, please let me know. I'm not sure what the best route to fixing the issue (if it is one) is yet, but I'll look into it in a bit.

VictorLamoine:
Happened to me already, I thought I was doing something wrong at that time!

gavanderhoorn:
This seems to be at least counter-intuitive, so I think this is not you (@Jmeyer1292) "using the system wrong". You could say that trajectories should be made up of segments that do not overlap, but that seems like an unreasonable restriction.

Perhaps the nr of points should be used to check whether a trajectory _could be_ completed at the point where `withinGoalConstraints(..)` returns `true`, but I'm not sure the `joint_trajectory_action` node has access to that information.

shaun-edwards:
Address in Indigo by #115

shaun-edwards:
This bug was not fixed by #115 or at least there was another manifestation with a different root cause.  In short, motions with the same begin and end point could return complete immediately if the `in_motion` status (and corresponding robot motion) did not begin immediately.  In other words, there exists a race condition between the setting and checking of the `in_motion` status.  More discussion on potential solutions can be found in #119.  As #119 is not a proper solution, this issue will remain open.

PR #119, opened 2015-07-17
Further Trajectory Completion Adjustments

Jmeyer1292:
This PR follows up on #115 addressing #114. 

While #115 removes the immediate check for goal completion, the action server can still terminate immediately upon the relieving its first feedback if a point toward the beginning of the trajectory is close to the end and the robot is slow to start.

This PR forces the action server to wait for half of the trajectories total duration before checking for the robot state against the end state. This gives time for the robot to start moving and publishing the `in motion` flag.

I have tested this pretty extensively in the Godel project and it seems to fix my issues. I am not sure if this is an appropriate solution, even as a temporary fix, but the issue remains and I thought I'd at least see what y'all think.

### DIFF
--- a/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
@@ -155,6 +155,12 @@ private:
   industrial_msgs::RobotStatusConstPtr last_robot_status_;
 
   /**
+   * \brief Time at which to start checking for completion of current 
+   * goal, if one is active
+   */
+  ros::Time time_to_check_;
+
+  /**
    * \brief The watchdog period (seconds)
    */
   static const double WATCHDOG_PERIOD_;// = 1.0;
--- a/industrial_robot_client/src/joint_trajectory_action.cpp
+++ b/industrial_robot_client/src/joint_trajectory_action.cpp
@@ -137,6 +137,8 @@ void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle & gh)
       gh.setAccepted();
       active_goal_ = gh;
       has_active_goal_ = true;
+      time_to_check_ = ros::Time::now() + 
+          ros::Duration(active_goal_.getGoal()->trajectory.points.back().time_from_start.toSec() / 2.0);
 
       ROS_INFO("Publishing trajectory");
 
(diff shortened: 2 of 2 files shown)


gavanderhoorn:
This to me seems like a work-around: why start checking after half of the trajectory has been executed? Why not only check after the last point of it has been sent?

Would this solve the problem of trying to execute a 'figure eight' kind of trajectory for instance? Where the path of the TCP intersects with the starting point multiple times?

shaun-edwards:
> This to me seems like a work-around: why start checking after half of the trajectory has been executed? Why not only check after the last point of it has been sent?

Ideally this would be how the problem is solved.  Unfortunately, the action interface does not know when the last point was sent.  Only the robot client node does.  A _better_ solution may be to add the action interface to the client node, but this is a larger, more invasive change.

> Would this solve the problem of trying to execute a 'figure eight' kind of trajectory for instance? Where the path of the TCP intersects with the starting point multiple times?

In this case, the `inMotion` status would prevent this confusion.  This PR solve the problem where a path begins and ends at that same point, combined with a delay in the start of the motion such that the `inMotion` status is `false` and the current robot position is simultaneously at the start and end points.

gavanderhoorn:
@shaun-edwards wrote:

> > This to me seems like a work-around: why start checking after half of the trajectory has been executed? Why not only check after the last point of it has been sent?
> 
> Ideally this would be how the problem is solved.  Unfortunately, the action interface does not know when the last point was sent.  Only the robot client node does.  A _better_ solution may be to add the action interface to the client node, but this is a larger, more invasive change.

Perhaps it would be an option to add a feedback signal between the action server and the robot client, in a similar way that actionlib uses `*Feedback` msgs to update its clients of status. It would make the bridging behaviour of the `industrial_robot_client` nodes more complete and would allow the action server to check for completion (actually only completion of the trajectory queueing, but that is a different issue).

> > Would this solve the problem of trying to execute a 'figure eight' kind of trajectory for instance? Where the path of the TCP intersects with the starting point multiple times?
> 
> In this case, the `inMotion` status would prevent this confusion.  This PR solve the problem where a path begins and ends at that same point, combined with a delay in the start of the motion such that the `inMotion` status is `false` and the current robot position is simultaneously at the start and end points.

Then this is still a work-around: checking `inMotion` is clearly not enough to be able to derive what the state of the robot / trajectory is. Something like the `EXECUTING` state as proposed in ros-industrial/rep#12 would allow us to unambiguously determine server state (especially since the sending of the trajectory points is only meant to queue them, and acks on those msgs are actually not a confirmation of trajectory execution).

gavanderhoorn:
Separating `inMotion` from something like `EXECUTING` also allows for trajectories in which there are segments with no motion. I can imagine welding would be one process where that would be a requirement.

shaun-edwards:
All of these suggestions are good ones.  I propose to accept this PR as is and log an issue for a better solution?  Thoughts?

gavanderhoorn:
Sure, we can merge this. At least as a workable work-around for now.

shaun-edwards:
+1 for the work-around.  Issue #114 was reopened to be addressed by a better, more appropriate fix later.  Thanks @Jmeyer1292 for the contribution and all for the discussion.

PR #121, opened 2015-10-22
Merging in released changes from Indigo

shaun-edwards:
DO NOT MERGE.  There are more changes than there should be.  I will have to track this down.


shaun-edwards:
This change was merged manually.  The discrepancies in the diff had to do with cherry picking changes and github comparing to the common ancestor (not the HEAD).

Issue #124, opened 2015-11-03
Joint trajectory action doesn't return complete in a timely manner


shaun-edwards:
The main discussion for this issue can be found here: https://github.com/ros-industrial/industrial_core/commit/cbdf113ec94e0780a1ffdd8f3d0ee3eef263c12c

simonschmeisser:
closed by #125

shaun-edwards:
Fixed in PR #125.

PR #125, opened 2015-11-30
Start checking for trajectory completion as soon as the first moving …

simonschmeisser:
…status arrives

This is a hotfix to #124 as discussed in cbdf113. It addresses cases where the computed duration of a trajectory execution is way longer than the time actually needed by the robot/controller

### DIFF
--- a/industrial_robot_client/src/joint_trajectory_action.cpp
+++ b/industrial_robot_client/src/joint_trajectory_action.cpp
@@ -45,7 +45,7 @@ const double JointTrajectoryAction::DEFAULT_GOAL_THRESHOLD_ = 0.01;
 JointTrajectoryAction::JointTrajectoryAction() :
     action_server_(node_, "joint_trajectory_action", boost::bind(&JointTrajectoryAction::goalCB, this, _1),
                    boost::bind(&JointTrajectoryAction::cancelCB, this, _1), false), has_active_goal_(false),
-                       controller_alive_(false)
+                       controller_alive_(false), has_moved_once_(false)
 {
   ros::NodeHandle pn("~");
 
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
fyi: failure of travis is caused by the repository not being configured properly for it. Nothing to do with this PR.

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -224,7 +226,7 @@ void JointTrajectoryAction::controllerStateCB(const control_msgs::FollowJointTra
     return;
   }
 
-  if (ros::Time::now() < time_to_check_)
+  if (!has_moved_once_ && (ros::Time::now() < time_to_check_))
```
shaun-edwards, line 229: `has_moved_once_` is only set if the robot status message is also published.  Not all robots publish this message, see [here](https://github.com/simonschmeisser/industrial_core/blob/indigo-devel/industrial_robot_client/src/joint_trajectory_action.cpp#L268).  I believe an additional check to `last_robot_status_`, like [here](https://github.com/simonschmeisser/industrial_core/blob/indigo-devel/industrial_robot_client/src/joint_trajectory_action.cpp#L241) would make this change compatible with older drivers.

simonschmeisser, line 229: well, even if has_moved_once_ is never set, the motion will succeed at either time_to_check if the robot is already at the goal or at same point later when it arrives there. So it is still compatible with those old drivers.

How would you do it in the old way? I don't know of any other information that could be used then.

shaun-edwards, line 229: You are correct...this will still work with the older drivers.

shaun-edwards:
+1...Thanks for the contribution @simonschmeisser.

PR #126, opened 2015-12-02
Add travis config as a git submodule from industrial_ci

130s:

PR #128, opened 2015-12-11
[CI] Switch from submodule to clone (see https://github.com/ros-industrial/industrial_ci/issues/3 thanks to @ipa-mdl)

130s:
With this change suggested by @ipa-mdl, maintainers no longer need to take any action when there's change in `industrial_ci` repo. See more in https://github.com/ros-industrial/industrial_ci/issues/3#issuecomment-164016120

Although as pointed out in https://github.com/ros-industrial/industrial_core/pull/126#issuecomment-161856876, the travis build is expected to fail, which should likely be unrelated to this change.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule ".ci_config"]
-	path = .ci_config
-	url = https://github.com/ros-industrial/industrial_ci.git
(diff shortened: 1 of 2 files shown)


shaun-edwards:
@130s, I don't see why the travis build fails.  The unit tests in `industrial_core` do test error conditions, but the tests ultimately passes (the code catches the error conditions as expected).

mathias-luedtke:
`catkin_test_results --verbose` will print a nice summary.
It perhaps the tests should be run with `-j1`  to not interleave the output.

shaun-edwards:
The failure is due to a roslaunch test in industrial_robot_simulator, see [here](https://travis-ci.org/ros-industrial/industrial_core/jobs/96340856#L4715).  +1 to merging, will address failure later.

PR #135, opened 2016-02-08
Removing git submodule for ci (no longer needed)

shaun-edwards:


shaun-edwards:
@130s, please review.

Note: test may fail on roslaunch-tests.  This seems to be a common issue with `industrial_ci`, but I haven't tracked it down yet.  The tests always fail on a missing dependency, but the `package.xml` clearly lists it.

shaun-edwards:
FYI...it appears the failure may be related to the work-around referenced by ros-industrial/ros_canopen#159

130s:
+1. Removing this directory shouldn't harm.

For the tests failure I'm keeping track at #129.

PR #136, opened 2016-02-09
Updated robot simulator to package.xml format 2.  Added test depends to workaround test failure

shaun-edwards:
Workaround for issue catkin/catkin_tools#285.  Should allow travis test jobs to succeed (or else not fail due to roslaunch test issues).

Note: This also required updating the package.xml format to version 2.

### DIFF
--- a/industrial_robot_simulator/package.xml
+++ b/industrial_robot_simulator/package.xml
@@ -21,4 +21,10 @@
   <exec_depend>rospy</exec_depend>
 
   <test_depend version_gte="1.9.55">roslaunch</test_depend>
+  <!--
+  industrial_robot_client is included as a work-around for:
+  https://github.com/catkin/catkin_tools/issues/285, once
+  this issue is fixed and released, it can be removed
+  -->
+  <test_depend>industrial_robot_client</test_depend>
 </package>
--- a/industrial_robot_simulator/package.xml
+++ b/industrial_robot_simulator/package.xml
@@ -1,4 +1,4 @@
-<package>
+<package format="2">
   <name>industrial_robot_simulator</name>
   <version>0.4.3</version>
   <description>The industrial robot simulator is a stand in for industrial robot driver node(s).  It adheres to the driver specification for industrial robot controllers.</description>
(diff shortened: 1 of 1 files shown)


shaun-edwards:
Tests now PASS!

@gavanderhoorn, can you please review updates to package.xml format 2?

gavanderhoorn:
Hm, I'm not too much of a fan of changing things just to accommodate a (temporary) bug in a build tool, but if this is needed, then so be it.

It would seem I'm partially responsible for us using `catkin_tools` instead of `catkin_make` anyway (ros-industrial/industrial_ci#3), so ..

industrial_robot_simulator/package.xml
```diff
@@ -9,18 +9,17 @@
...
-  <run_depend>industrial_robot_client</run_depend>
-  <run_depend>industrial_msgs</run_depend>
-  <run_depend>python-rospkg</run_depend>
-  <run_depend>rospy</run_depend>
+
+  <depend>std_msgs</depend>
+  <depend>sensor_msgs</depend>
+  <depend>control_msgs</depend>
+  <depend>trajectory_msgs</depend>
+  <depend>industrial_msgs</depend>
```
gavanderhoorn, line 14: It is recommended to sort all dependencies alphabetically.

industrial_robot_simulator/package.xml
```diff
@@ -9,18 +9,17 @@
...
+
+  <exec_depend>industrial_robot_client</exec_depend>
+  <exec_depend>python-rospkg</exec_depend>
+  <exec_depend>rospy</exec_depend>
```
gavanderhoorn, line 21: same here (and below).

gavanderhoorn:
+1 on the changes (ie: they are syntactically correct), but -1 for why we are doing this. As @ipa-mdl already commented in ros-industrial/ros_canopen#155, this can / should be reverted once `catkin_tools` 0.4.0 is released. I'd rather we switch to `catkin_make` for the time being (don't know how feasible that is though, @130s?).

mathias-luedtke:
> Note: This also required updating the package.xml format to version 2.

Actually this is an improvement.

It should work with format 1 as well, if you add `industrial_robot_client` as a `build_depend`.
(It feels odd, but it enforces the correct order)

industrial_robot_simulator/package.xml
```diff
@@ -9,18 +9,17 @@
...
+
+  <test_depend version_gte="1.9.55">roslaunch</test_depend>
+  <test_depend>industrial_robot_client</test_depend>
```
mathias-luedtke, line 29: This line should be a separate commit. This simplifies the revert.

shaun-edwards:
I split the commits (one for upgrade to format 2, one for the workaround).  I sorted the depends alphabetically.  

I'm not excited about committing a work-around, so I added a comment about the reason for the `test_depends`.

@gavanderhoorn, ok to merge?

gavanderhoorn:
@shaun-edwards: looks like Travis test failed, despite the work-around?

shaun-edwards:
Looks like it may be related to the number of parallel jobs, see [here](https://docs.travis-ci.com/user/customizing-the-build/#Limiting-Concurrent-Builds).  I've set the number of parallel jobs to 1.  This should solve the problem of multiple tests using the same socket port.

shaun-edwards:
@gavanderhoorn, all fixed now...ok to merge?

gavanderhoorn:
+1 then.

But lets log an issue to revert 40fa0da once `catkin_tools` has been fixed.

PR #140, opened 2016-03-15
Make industrial_robot_simulator python3 compatible.

de-vri-es:
This PR makes two minor changes to make the industrial robot simulator compatible with python3. There should be no functional changes at all for python 2.

I only had to swap `operator <>` with `operator !=` and import the `Queue` as `queue` in python 3.

### DIFF
--- a/industrial_robot_simulator/industrial_robot_simulator
+++ b/industrial_robot_simulator/industrial_robot_simulator
@@ -34,7 +34,12 @@
 import rospy
 import copy
 import threading
-import Queue
+
+import sys
+if sys.version_info.major == 2:
+    import Queue
+else:
+    import queue as Queue
 
 # Publish
 from sensor_msgs.msg import JointState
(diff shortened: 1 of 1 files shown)


industrial_robot_simulator/industrial_robot_simulator
```diff
@@ -34,7 +34,12 @@
 import rospy
 import copy
 import threading
-import Queue
+
+import sys
+if sys.version_info[0] == 2:
+	import Queue
+else:
+	import queue as Queue
```
gavanderhoorn: Could you please change indentation to spaces?

industrial_robot_simulator/industrial_robot_simulator
```diff
@@ -34,7 +34,12 @@
 import rospy
 import copy
 import threading
-import Queue
+
+import sys
+if sys.version_info[0] == 2:
```
gavanderhoorn: Suggestion: use `sys.version_info.major` instead: intent is immediately clear and no magic nrs.

industrial_robot_simulator/industrial_robot_simulator
de-vri-es: Good point, updated.

industrial_robot_simulator/industrial_robot_simulator
de-vri-es: Agreed, changed.

de-vri-es:
Any other thoughts?

shaun-edwards:
+1... @de-vri-es, thanks so much for the contirbution.

Issue #130, opened 2016-01-11
Typo in name of enum value in JointTrajPt::SpecialSeqValue


gavanderhoorn:
In both `joint_traj_pt.h` ([here](https://github.com/ros-industrial/industrial_core/blob/12a74a1f9f26aea0ee075edaf4c84473bd8e112a/simple_message/include/simple_message/joint_traj_pt.h#L52-L59)) and `joint_traj_pt_full.h` ([here](https://github.com/ros-industrial/industrial_core/blob/12a74a1f9f26aea0ee075edaf4c84473bd8e112a/simple_message/include/simple_message/joint_traj_pt_full.h#L52-L59)) there is the following typo:

``` c++
enum SpecialSeqValue
{
  START_TRAJECTORY_DOWNLOAD = -1,
  START_TRAJECOTRY_STREAMING = -2,
  END_TRAJECTORY = -3,
  STOP_TRAJECTORY = -4
};
```

Note the _TRAJECOTRY_.

I don't believe anything but `STOP_TRAJECTORY` is ever used, but the typo should still be fixed imho.

Seeing as this code has been released into LTS distributions, we probably need to keep the incorrect enum value around as well.

PR #142, opened 2016-04-08
Fix for issue #130: correct typo in SpecialSeqVal enum member name

gavanderhoorn:
As per subject.

c5add37 fixes the typo (adds an additional enum member with corrected spelling) and marks the old one deprecated (comment only).

Added some Doxygen comments in 6871959.


shaun-edwards:
+1...thanks for the contribution.

PR #143, opened 2016-05-11
industrial_robot_client: Fix signature of goal and cancel callbacks.

de-vri-es:
The goal and cancel callbacks of `ActionServer` take `GoalHandle`s by value [1], but the callbacks defined in `JointTrajectoryActionServer` took them by reference. I'm not sure why this compiled before C++11, but it doesn't now. Since g++ 6 compiles with C++11 support enabled by default, it is quite important to fix this.

Note that `GoalHandle` already acts like a reference type; it contains only (shared) pointers. So accepting them by value or by reference wont change the behaviour of the program.

[1] https://github.com/ros/actionlib/blob/da39083c513066e0d3d05346f3b1be9f9f90268d/include/actionlib/server/action_server.h#L87

### DIFF
--- a/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
@@ -185,7 +185,7 @@ private:
    * \param gh goal handle
    *
    */
-  void goalCB(JointTractoryActionServer::GoalHandle & gh);
+  void goalCB(JointTractoryActionServer::GoalHandle gh);
 
   /**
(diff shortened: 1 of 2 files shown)


shaun-edwards:
@de-vri-es, Thank you for the contribution.  +1 to merging.  I am curious as to why you are using gc++ 6?  Does it come standard with a version of Ubuntu?  I wonder if we should add this configuration to our travis CI testing?

de-vri-es:
I'm running Arch Linux and compiling from source. Arch recently switched to gcc 6.0. Painfully paving the way for other distros :]

PR #144, opened 2016-05-31
[jade] Update Travis config for Jade

130s:
Current config in `jade` branch is for Indigo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,13 @@ notifications:
 #      - jane@doe
 env:
   matrix:
-    - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
-    - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
     - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
     - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - ROS_DISTRO=kinetic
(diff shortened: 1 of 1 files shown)


130s:
Also added is `Kinetic` support as an option.

shaun-edwards:
@130s, I'm interested in your thoughts here.  The`jade` and `indigo` branches are the stable branches of the `indigo-devel` branch.  At some point I expect to create a `jade-devel` for jade specific changes/improvements.  At which point, the `jade` branch will follow that.  My question is whether the each branch should be built for their specific ROS version or if I should build for all versions as suggested in this PR?

130s:
> My question is whether the each branch should be built for their specific ROS version or if I should build for all versions as suggested in this PR?

I now see the purpose of branches. I updated the commit so that we only check `jade` branch. 

And also I keep kinetic build assuming for the future that this branch may be escalated to `kinetic` branch?

shaun-edwards:
Jade on `shadow-fixed` failed...odd?  I'll have to take a look.

130s:
@shaun-edwards 

> Jade on shadow-fixed failed...odd? I'll have to take a look.

https://travis-ci.org/ros-industrial/industrial_core/jobs/134473485#L2777

> [industrial_trajectory_filters:cmake]   Could not find a package configuration file provided by
> [industrial_trajectory_filters:cmake]   "moveit_ros_planning" with any of the following names:

This is because `moveit_ros` for jade is [not available in shado-fixed](http://repositories.ros.org/status_page/ros_jade_default.html?q=moveit_ros).

shaun-edwards:
@130s, what's your recommendation here?  Should the `shadow-fixed` builds be allowed to fail and/or should we just accept this as is, knowing what is causing the failure.

130s:
That's my best guess..(moveit for Jade might still take some time to become available, so does Kinetic). I've updated the commit. We'll see how it works.

130s:
Test passed.

shaun-edwards:
+1... @130s, thanks for the contribution.

PR #150, opened 2016-09-09
Use urdf::*ShredPtr instead of boost::shared_ptr

jspricke:
urdfdom_headers uses C++ std::shared_ptr. As it exports it as custom
*SharedPtr type, we can use the to stay compatible.

### DIFF
--- a/industrial_utils/CMakeLists.txt
+++ b/industrial_utils/CMakeLists.txt
@@ -6,14 +6,19 @@ find_package(catkin REQUIRED COMPONENTS roscpp urdf)
 
 find_package(Boost REQUIRED COMPONENTS system)
 
+find_package(urdfdom_headers 0.4 REQUIRED)
+
 catkin_package(
     CATKIN_DEPENDS roscpp urdf
     INCLUDE_DIRS include
     LIBRARIES ${PROJECT_NAME}
+    DEPENDS urdfdom_headers
 )
 
 include_directories(include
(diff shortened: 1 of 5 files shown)


jspricke:
Note that the CI fails because urdfdom_headers is installed with version 0.2.3, whereas at least 0.4 is required.

gavanderhoorn:
Then I'm guessing that this also shouldn't be merged into `indigo-devel`, as Indigo doesn't have `urdfdom_headers` > 0.2.3?

jspricke:
Right, can you open a new branch? ;)

shaun-edwards:
I pulled these changes into my own repo and created a new PR #166.  The CI test for jade fail, but kinetic passes: https://travis-ci.org/ros-industrial/industrial_core/builds/192262495

Should this PR build under jade?

gavanderhoorn:
Well, at least `urdfdom_headers` > `0.2.3` (ie: `0.4+`) is only available on Xenial and up (at least on Ubuntu, see [here](https://launchpad.net/ubuntu/+source/urdfdom-headers)).

shaun-edwards:
Ok...unless some objects ( @jspricke  ), I'll pull this change into the `kinetic-devel` branch.

jspricke:
Oups, completely forgot about this one, sorry. Actually we added compatibility headers, so the dependency on 0.4 is not needed anymore. Please don't merge, I will try to update it later.

PR #151, opened 2016-10-01
robot_simulator: clarify err msg when controller_joint_names is missing.

gavanderhoorn:
As per subject.


shaun-edwards (approved):
All good!

shaun-edwards:
Thanks for the contribution....merging.

PR #153, opened 2016-10-17
robot_status: missing reply to SERVICE_REQUEST. 

MariniAlberto:
Fix in robot_status_message and relay_handler.

Wrong condition check at line 83 of robot_status_relay_handler.cpp (comparing CommType with MessageType).
In order to fix this issue, I had to modify also the robot_status_message class, adding get\* and set\* methods for the comm_type attribute.

### DIFF
--- a/simple_message/include/simple_message/messages/robot_status_message.h
+++ b/simple_message/include/simple_message/messages/robot_status_message.h
@@ -110,8 +110,38 @@ public:
   {
     return this->status_.byteLength();
   }
+  
+  /**
+   * \brief Gets the communication type of the message
+   * 
+   * \return the value of the comm_type parameter (refer to simple_message::CommTypes::CommType)
+   */
+  int getCommType() const
+  {
+    return comm_type_;
+  }
+  
 
   industrial::robot_status::RobotStatus status_;
+  
+protected:
+
+  /**
+   * \brief Sets the communication type of the message
+   *
+   * \param comm_type: value of the comm_type parameter (refer to simple_message::CommTypes::CommType)
+   */
+  void setCommType(int comm_type = industrial::simple_message::CommTypes::INVALID)
+  {
+    this->comm_type_ = comm_type;
+  }
+  
+private:
+  
+  /**
+   * \brief Communications type (see simple_message::CommTypes::CommType)
+   */
+  industrial::shared_types::shared_int comm_type_;
 
 
 };
--- a/simple_message/include/simple_message/messages/robot_status_message.h
+++ b/simple_message/include/simple_message/messages/robot_status_message.h
@@ -111,38 +111,8 @@ public:
     return this->status_.byteLength();
   }
   
-  /**
-   * \brief Gets the communication type of the message
-   * 
-   * \return the value of the comm_type parameter (refer to simple_message::CommTypes::CommType)
-   */
-  int getCommType() const
-  {
-    return comm_type_;
-  }
-  
 
   industrial::robot_status::RobotStatus status_;
-  
-protected:
-
-  /**
-   * \brief Sets the communication type of the message
-   *
-   * \param comm_type: value of the comm_type parameter (refer to simple_message::CommTypes::CommType)
-   */
-  void setCommType(int comm_type = industrial::simple_message::CommTypes::INVALID)
-  {
-    this->comm_type_ = comm_type;
-  }
-  
-private:
-  
-  /**
-   * \brief Communications type (see simple_message::CommTypes::CommType)
-   */
-  industrial::shared_types::shared_int comm_type_;
-
 
 };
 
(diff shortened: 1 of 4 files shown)


shaun-edwards:
@MariniAlberto thanks for the PR.  This indeed looks like a bug.  The travis job seems to be failing due to socket issues, so for now we will ignore this.  Just a couple of items to review.

simple_message/include/simple_message/messages/robot_status_message.h
```diff
@@ -110,8 +110,38 @@ class RobotStatusMessage : public industrial::typed_message::TypedMessage
   {
     return this->status_.byteLength();
   }
+  
+  /**
+   * \brief Gets the communication type of the message
+   * 
+   * \return the value of the comm_type parameter (refer to simple_message::CommTypes::CommType)
+   */
+  int getCommType() const
```
shaun-edwards: Is there a reason you added this method here instead of in the base class [TypedMessage](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/simple_message/include/simple_message/typed_message.h)?  `TypedMessage` implements [getMessageType](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/simple_message/include/simple_message/typed_message.h#L137) here.

simple_message/include/simple_message/messages/robot_status_message.h
```diff
@@ -110,8 +110,38 @@ class RobotStatusMessage : public industrial::typed_message::TypedMessage
...
   industrial::robot_status::RobotStatus status_;
+  
+protected:
+
+  /**
+   * \brief Sets the communication type of the message
+   *
+   * \param comm_type: value of the comm_type parameter (refer to simple_message::CommTypes::CommType)
+   */
+  void setCommType(int comm_type = industrial::simple_message::CommTypes::INVALID)
```
shaun-edwards: See comment above.

simple_message/src/messages/robot_status_message.cpp
```diff
@@ -66,6 +66,7 @@ bool RobotStatusMessage::init(industrial::simple_message::SimpleMessage & msg)
   bool rtn = false;
   ByteArray data = msg.getData();
   this->init();
+  this->setCommType(msg.getCommType());
```
shaun-edwards, line 69: This is more of a style issue.  Using accessors in a class function is not necessary (although fine) since these classes are relatively simple.  Not required to fix, just my opinion.

shaun-edwards (changes requested):
Just a few items to fix/discuss.  In the end, I think we will pull this into `indigo-devel` and then push it out to both the `indigo` and `jade` branches.  The non-`devel` branches are release branches.

simple_message/include/simple_message/messages/robot_status_message.h
MariniAlberto: It could be placed also in the TypedMessage class. I just didn't want to mess up with a base class from which other classes are inherited.

simple_message/include/simple_message/messages/robot_status_message.h
MariniAlberto: The same as above.

simple_message/src/messages/robot_status_message.cpp
MariniAlberto: Sorry but I don't understand what you mean. Maybe I'm wrong, but how can I retrieve the comm_type_ value (private defined) without using those methods?

simple_message/include/simple_message/messages/robot_status_message.h
gavanderhoorn: This is most likely missing functionality in other req/reply classes as well, so it makes sense to add this to shared parts.

simple_message/src/messages/robot_status_message.cpp
shaun-edwards: You are right...my bad.

simple_message/include/simple_message/messages/robot_status_message.h
shaun-edwards: @MariniAlberto, please move the method down into TypedMessage and then I'm ready to approve.  Thanks for the contribution.

simple_message/include/simple_message/messages/robot_status_message.h
shaun-edwards: Move into TypedMessage.

simple_message/include/simple_message/messages/robot_status_message.h
MariniAlberto: @shaun-edwards , I moved the methods into TypedMessage class and made the push on this branch. Please check if everything's ok. Thanks.

shaun-edwards (approved):

Issue #149, opened 2016-09-05
skipped tests make build status: unstable


gavanderhoorn:
See [build 4](http://build.ros.org/job/Idev__industrial_core__ubuntu_trusty_amd64/4/testReport/) fi, but the general test result trend is also [unstable](http://build.ros.org/job/Idev__industrial_core__ubuntu_trusty_amd64/).

gavanderhoorn:
@shaun-edwards: I'm assuming you disabled those tests? Is there anything we can fix, so we can re-enable them?

shaun-edwards:
The tests are disabled because they need roscore to execute.  The "right" thing to do create a "rostest" to run.  Now that I've done this a few times, it should be easy.

shaun-edwards:
Back to stable: http://build.ros.org/job/Jdev__industrial_core__ubuntu_trusty_amd64/6/

gomezc:
I'll be looking at this for wrid18

gavanderhoorn:
@gomezc: were you able to make any progress with this?

gomezc:
I made some progress but was not able to complete it and have been distracted since. If no one else is going to take it, I will still work on it in my spare time, but if someone else is looking at it, I will gladly relinquish. :)

gavanderhoorn:
Could you at least describe what you did / what you found out? If/when others take this up, that may save them time.

gavanderhoorn:
It would still be nice if we could get a summary of what progress was made, but seeing as it's already been a year, that'd probably be difficult.

PR #162, opened 2017-01-05
Temporary fix, commented out disabled test to remove unstable build status

shaun-edwards:
This is a temporary fix for #149.  I commented out the test instead of disabling it.  The end result is the same, the test is not executed.  This should clear up the unstable build issue.


shaun-edwards:
I'm going to merge this without review for the sake of being able to release.  The code change is small and only affects non-critical tests.  Addressing #149 is the real solution to this change.

PR #166, opened 2017-01-16
PR to replace #150 (pulled into jade-devel branch)

shaun-edwards:


shaun-edwards:
See discussion in #150

PR #167, opened 2017-01-16
Updated travis config for kinetic distro

shaun-edwards:


gavanderhoorn:
I know I do this myself now and then, but no reviews? :)

shaun-edwards:
I'm in hurry to get a kinetic release out.  Swri has requested we release a
kinetic version.  The rest of this week is going to be hard.  I figured it
was a low enough risk.

PR #170, opened 2017-01-17
Use urdf::*SharedPtr instead of boost::shared_ptr

jspricke:
urdfdom_headers uses C++ std::shared_ptr. As it exports it as a custom
*SharedPtr type, we can use them to stay compatible.

### DIFF
--- a/industrial_utils/include/industrial_utils/utils.h
+++ b/industrial_utils/include/industrial_utils/utils.h
@@ -75,7 +75,7 @@ bool isSame(const std::vector<std::string> & lhs, const std::vector<std::string>
  *
  * \return true if successful, false if error occurred (e.g. branching tree)
  */
-bool findChainJointNames(const boost::shared_ptr<const urdf::Link> &link, bool ignore_fixed,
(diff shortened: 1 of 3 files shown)


jspricke:
Replacement for #150 for kinetic-devel

gavanderhoorn:
Wouldn't this now also work under Indigo & Jade? Because of the compatibility header?

Personally I'd like to see it in Indigo and Jade then as well.

jspricke:
Yes, should be fine for all.

jspricke:
should I send pull requests for them as well?

gavanderhoorn:
I'll change the target for this PR to `indigo-devel` after Travis has completed. I think we can manage the other PRs ourselves.

shaun-edwards:
@gavanderhoorn, do you want me to handle this?

shaun-edwards (approved):

VictorLamoine:
@shaun-edwards CI is happy with this PR, you can merge

shaun-edwards:
@VictorLamoine, sorry for the delay.  Github wouldn't let me merge this a couple days back...but now it magically allowed me.

PR #174, opened 2017-03-07
Added a disable/enable service that stops the industrial_robot_client from accepting any incoming goals

pbeeson:
For safety, it is sometimes desirable to have a "software level e-stop".  This essentially does that for ROS-I robot by adding a service that "pauses" or "unpauses" the robot by allowing goals to come through.  If the robot is paused, any current goal is canceled via the cancelCB() function, stopping the robot.

Discussed here: https://github.com/ros-industrial/motoman/issues/152


gavanderhoorn:
Thanks for the PR.

Something to think about: semantically I would say that a "paused" robot can be resumed, which would imply it would also resume whatever it was doing. As the `pause_robot` service cancels any active goal, this is not the case with this implementation.

I'm fine with cancelling the current goal, but then I feel the service should not be called `pause_*`.

pbeeson:
I agree, and don't mind changing this to "Halted"  or "E-Stopped" or similar.

pbeeson:
Terminology has been changed from pause/unpause to disable/enable.

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -77,6 +80,29 @@ void JointTrajectoryAction::robotStatusCB(const industrial_msgs::RobotStatusCons
...
+  else
+    res.message="Robot is now ";
+
+  if (disabled_)
+    res.message += "disabled";
+  else res.message += "enabled";
+  
+  return true;
+  
+}
```
gavanderhoorn, line 109: This is again more of a question than a request (so feel free to disagree / tell me to do this myself), but would it be nice to split this into a `disable_robot` and an `enable_robot` service (and use `std_srvs/Trigger` for the type)?

It would simplify this code and be very explicit from a component interface perspective.

Finally: should a getter service be added as well? That way we can retrieve the current state.

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -123,6 +149,14 @@ void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle gh)
     return;
   }
 
+  if (disabled_)
+  {
+    control_msgs::FollowJointTrajectoryResult rslt;
+    rslt.error_code = control_msgs::FollowJointTrajectoryResult::INVALID_GOAL;
+    gh.setRejected(rslt, "Robot has been disabled.  Goals will be ignored until the service is called to enable the robot.");
+    return;
+  }
```
gavanderhoorn, line 193: Could a `ROS_WARN(..)` be added here similar to how this is done below? I can already imagine the ROS Answers questions "my goals keep getting ignored, why isn't my robot moving !!11~~" from users not checking actionlib goal statuses.

gavanderhoorn:
We should also update the [robot driver spec](http://wiki.ros.org/Industrial/Industrial_Robot_Driver_Spec) page after we merge this (section 1.3.3, *services*).

pbeeson:
@gavanderhoorn I addressed all your issues, as they were reasonable.

Now there are Trigger services for Enagle and Disable and a GetEnabledStatus message in industrial_msgs with an appropriate service.  This is also in motoman_driver as desired.  Maybe someone can consider testing this PR?  It works for us.

shaun-edwards:
Waiting on discussion in ros-industrial/motoman#155

pbeeson:
So, I realize this was similar to the PR in the motoman driver, but this is at the industrial client level so it allows users to use and test safe control enable/disable  in industrial sim and any other ROS industrial robot at a "mid-level" of the control flow.

industrial_robot_client/src/joint_trajectory_action.cpp
shaun-edwards: @pbeeson, could you address this request?

industrial_robot_client/src/joint_trajectory_streamer.cpp
```diff
@@ -69,19 +69,20 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
...
+  
   ROS_DEBUG("Current state is: %d", state);
+
   if (TransferStates::IDLE != state)
   {
-    if (msg->points.empty())
-      ROS_INFO("Empty trajectory received, canceling current trajectory");
-    else
-      ROS_ERROR("Trajectory splicing not yet implemented, stopping current motion.");
+    // if (msg->points.empty())
```
shaun-edwards: Is there a reason that the you commented out these messages?

industrial_robot_client/src/joint_trajectory_streamer.cpp
```diff
@@ -167,7 +168,7 @@ void JointTrajectoryStreamer::streamingThread()
     switch (this->state_)
     {
       case TransferStates::IDLE:
-        ros::Duration(0.250).sleep();  //  slower loop while waiting for new trajectory
+        ros::Duration(0.1).sleep();  //  slower loop while waiting for new trajectory
```
shaun-edwards: No a big deal, but I'm curious why you needed/wanted a faster loop.

industrial_robot_client/src/joint_trajectory_streamer.cpp
pbeeson: This was probably because the final client considered any red ERROR message they saw on the console as a reportable event.  I didn’t realize this had made it into the top level.

industrial_robot_client/src/joint_trajectory_streamer.cpp
pbeeson: At some point we were having timing issues with the Motoman driver and this was probably done on site in that debug.  Again, I apologize for having Croft like this in the PR.  I can submit a new one with only the relevant changes if that makes your life easier.

pbeeson: Some of this cruft may have happened after the original PR was submitted, but yeah, that’s what feature branches are for.  Mid-2017 was hectic so I can spend some time cleaning this up properly if needed.

pbeeson: I recall now.  In our production cell we send probably 30 trajectories per task cycle.  This is a time critical application with a hard 60 second cycle time.  A 0.25 loop meant a possible worst-case delay of 7.5s over the 30 trajectories.  7.5s versus 3s (at 0.1 s loop) bought us 4.5s per cycle best case.  In practice it was less but still noticeable.

industrial_robot_client/src/joint_trajectory_streamer.cpp
pbeeson: I fixed this in the current version of the PR.

industrial_robot_client/src/joint_trajectory_action.cpp
pbeeson: Yep.  Done.

industrial_robot_client/src/joint_trajectory_action.cpp
pbeeson: I guess I did this split a long time ago.

pbeeson: Yes.  It has been split and there is a getter.

pbeeson:
Closing this PR because I made it from my main branch a year and a half ago and no that branch has other changes that should be separate PRs.  PR #215 has the minimal commits that address this PR.

Issue #157, opened 2016-12-07
multiple: CMake targets depend on non-existent targets


gavanderhoorn:
Seen in various Travis builds:

```
CMake Warning (dev) at /opt/ros/indigo/share/catkin/cmake/test/tests.cmake:137 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "joint_trajectory_action" of target
  "run_tests_industrial_robot_simulator_roslaunch-check_launch" does not
  exist.
Call Stack (most recent call first):
  /opt/ros/indigo/share/roslaunch/cmake/roslaunch-extras.cmake:51 (catkin_run_tests_target)
  CMakeLists.txt:18 (roslaunch_add_file_check)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at /opt/ros/indigo/share/catkin/cmake/test/tests.cmake:149 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "joint_trajectory_action" of target
  "_run_tests_industrial_robot_simulator_roslaunch-check_launch" does not
  exist.
Call Stack (most recent call first):
  /opt/ros/indigo/share/roslaunch/cmake/roslaunch-extras.cmake:51 (catkin_run_tests_target)
  CMakeLists.txt:18 (roslaunch_add_file_check)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

gavanderhoorn:
[this](https://travis-ci.org/ros-industrial/industrial_core/jobs/181766807) is one of those builds (search for `error`).

PR #185, opened 2017-10-07
Fix for issue #157: don't depend on non-existing targets in ind_rob_sim build file

gavanderhoorn:
`joint_trajectory_action` is not a target as far as CMake is concerned at that point in the build. So we can't depend on it.

This is similar to https://github.com/ros-industrial/fanuc/pull/187.


gavanderhoorn:
This should be backported to `indigo-devel` as well.

shaun-edwards (approved):

PR #189, opened 2018-02-14
Using the 'doc' attribute on 'arg' elements.

hsd-dev:
add doc strings to all launch file args.


shaun-edwards (approved):

shaun-edwards:
Thanks for the contribution.

Issue #188, opened 2018-02-08
`simple_message` doesn't compile without `LINUXSOCKETS` defined


jdlangs:
This is mainly a problem when a dependent project includes some header files from `simple_message` and the developer doesn't realize he/she has to set this compile flag. You end up getting errors about socket types not defined and the only way to figure out what's going on is to go look in the code.

Should this be changed so that the `LINUXSOCKETS` behavior is default and the other platform builds require a flag?

shaun-edwards:
It's been a while, but this sounds like an issue I've had before.  It's a lack of CMake ability on my part.  Do you know what change to make?

jdlangs:
I'm not sure what the best practices around conditional compilation for different platforms are. In this case, since it looks like only other condition is the the `MOTOPLUS` flag, would it make sense just to change `#ifdef LINUXSOCKETS` to `#ifndef MOTOPLUS`?

shaun-edwards:
That might work.  It wouldn't support more than two options though (like a third socket library).  Still feels like a CMake approach would be better, like exporting flags or something.  I'd accept a PR on this one.  Very few people (or maybe no one) use the headers for MOTOPLUS (shared code was thrown out a long time ago).

gavanderhoorn:
@shaun-edwards wrote:
> Still feels like a CMake approach would be better, like exporting flags or something. 

exactly, that would be the correct way to do this, although we'd have to make a choice for a particular set of value(s) for the `define`s.

For Catkin packages, this could be done with a `CFG_EXTRAS` file. See [export gcc flags though catkin? CFG_EXTRAS?](https://answers.ros.org/question/173947) on ROS Answers for an example (discusses compiler flags, but it works for other things as well).

An alternative (and non-Catkin) approach would be to make sure a `simple_message_DEFINITIONS` variable gets set whenever `simple_message` is `find_package(..)`d. Consumers would then need to add something like this to their `CMakeLists.txt` (newer CMake versions might use slightly different syntax, I haven't checked):

```cmake
add_definitions(${simple_message_DEFINITIONS})
```

Such a variable should probably be part of the `simple_messageConfig.cmake` file (a "config file" in CMake parlance). Not sure how to implement that with Catkin.

[...]

Issue #65, opened 2014-01-10
simple_message: possibility for name clashes with defines (INT32, INT64, ..)


gavanderhoorn:
We should probably namespace all defines used in the `simple_message` package, as they are rather generically named right now. Especially `INT32`, `INT64`, `ROS` and `LINUXSOCKETS` have the potential to clash with defines from other headers / libraries.

I propose to prefix them with something like `SIMPLE_MESSAGE_`.

Now that we start looking into properly exporting those defines to users of the package as well, this could become a real issue.

PR #190, opened 2018-02-21
Export platform build flags to dependent catkin projects

jdlangs:
This fixes the problem from #188 by using the `CFG_EXTRAS` option of the `catkin_package` command to export default compiler definitions to all projects which depend on `simple_message`. It also addresses a few of the names mentioned in #65 by adding a `SIMPLE_MESSAGE_` prefix to the definitions.

The "documentation" of these definitions for package users is right now only in `CMakeLists.txt` and the new `platform_build_flags.cmake` file. Let me know if there is anywhere else it would be appropriate to note this behavior.

### DIFF
--- a/simple_message/CMakeLists.txt
+++ b/simple_message/CMakeLists.txt
@@ -9,11 +9,11 @@ set(ROS_BUILD_STATIC_LIBS true)
 set(ROS_BUILD_SHARED_LIBS false)
 
 # The simple_message library is designed to cross compile on Ubuntu
-# and various robot controllers.  This requires conditionally compiling
-# certain functions and headers.  The definition below enables compiling
-# for a ROS node.
-add_definitions(-DROS=1)           #build using ROS libraries
-add_definitions(-DLINUXSOCKETS=1)  #use linux sockets for communication
+# and various robot controllers. This requires conditionally compiling
+# certain functions and headers. The default flags in the file included
+# here enable compiling for a ROS node. This file is also exported to
+# dependent packages via the `catkin_package` commmand.
+include(cmake/platform_build_flags.cmake)
 
 set(SRC_FILES src/byte_array.cpp
 	src/simple_message.cpp
@@ -71,7 +71,7 @@ catkin_package(
     CATKIN_DEPENDS roscpp industrial_msgs
     INCLUDE_DIRS include
     LIBRARIES ${PROJECT_NAME}_dummy
-    CFG_EXTRAS issue46_workaround.cmake
+    CFG_EXTRAS platform_build_flags.cmake issue46_workaround.cmake
 )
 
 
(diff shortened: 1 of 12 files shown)


gavanderhoorn:
Thanks for the PR, this looks like a useful change.

I'll submit a review shortly with some comments.

gavanderhoorn:
As it's just a few questions/comments really, let's not start a review yet.

 1. it seems your commit isn't attributed to your account, you may want to fix that
 1. did you remove the `CFG_EXTRAS` entry from `industrial_robot_client` on purpose? Is it no longer needed?
 1. addressing #65 is nice, but we have to be careful with this change so as to not introduce any breaking changes in Kinetic (which is an LTS). Could I ask you to refactor the changes that introduce the prefix into a separate commit? We should then also keep the old defines in Kinetic - and provide the new ones as well - add a deprecation notice and then in Lunar we can remove the old ones.

simple_message/CMakeLists.txt
```diff
@@ -9,11 +9,11 @@ set(ROS_BUILD_STATIC_LIBS true)
...
-# and various robot controllers.  This requires conditionally compiling
-# certain functions and headers.  The definition below enables compiling
-# for a ROS node.
-add_definitions(-DROS=1)           #build using ROS libraries
-add_definitions(-DLINUXSOCKETS=1)  #use linux sockets for communication
+# and various robot controllers. This requires conditionally compiling
+# certain functions and headers. The default flags in the file included
+# here enable compiling for a ROS node. This file is also exported to
+# dependent packages via the `catkin_package` commmand.
+include(cmake/platform_build_flags.cmake)
```
jrgnicho, line 16: I think documenting this as a subsection in the README might be helpful

jdlangs:
@gavanderhoorn sorry it took me a little while to come back to this. Responding to your comments:

1. Fixed
2. It does not appear to be needed. I tested building against the `industrial_robot_client` library and it only breaks if I remove the `issue46_workaround` file from the `CFG_EXTRAS` in `simple_message`.
3. I've put the definition renaming in a separate commit. Is there a better deprecation strategy than just continuing to provide the old defines with a note to remove them later? I couldn't see a way to keep them and emit a warning in the user's code.

simple_message/CMakeLists.txt
jdlangs: I guess the [wiki page](http://wiki.ros.org/simple_message) would be best but it may be out-of-date (last edit ~3 years ago)

jdlangs:
Refreshed this PR for ROS-I day. This is unchanged from last year but I believe everything brought up had been addressed.

jdlangs:
@jrgnicho would you want to take another look since you commented the first time around?

gavanderhoorn:
Replacement/follow-up in #262.

thanks for the initial PR @jdlangs :+1:

PR #193, opened 2018-07-11
Update the libraries install tags - follow the official documentation

ipa-nhg:


mathias-luedtke:
This is not really needed.
Please include the dummy lib as well.

mathias-luedtke:
It might be even better to get rid of these dummy libs at all ;)

gavanderhoorn:
Thanks for the PR @ipa-nhg.

re: `dummy_lib`: that is already installed [here](https://github.com/ros-industrial/industrial_core/blob/e3fac7f5018c6c8db9e859e89947e2b52159e0da/simple_message/CMakeLists.txt#L93). Could you please remove it from the lines you updated *or* remove the `install(..)` statement that installs the dummy lib?

ipa-nhg:
@gavanderhoorn thanks for the review, just remove this commit https://github.com/ros-industrial/industrial_core/pull/193/commits/04ee84e384d7fb20df3701a351baf87124cffa04 should be ok. I will rebase and drop it

gavanderhoorn:
You just added 66e24525, was that intentional?

ipa-nhg:
@gavanderhoorn no, I removed a previous commit. Probably the branch was updated (using the web interface "update branch" button yesterday??)....

gavanderhoorn:
I don't know, I don't see any updates in the gh ui.

It would be good if we can keep these things separate. Could you drop 66e2452 from your `LibInstallTags` branch?

ipa-nhg:
@gavanderhoorn Yep! done!

gavanderhoorn (approved):
Looks good.

gavanderhoorn:
Thanks again @ipa-nhg :+1:

Issue #108, opened 2015-03-04
simple_message: gethostbyname() is deprecated / obsolete


gavanderhoorn:
According to [man 3 gethostbyname](http://man7.org/linux/man-pages/man3/gethostbyname.3.html):

> The `gethostbyname*()`, `gethostbyaddr*()`, `herror()`, and `hstrerror()` functions are obsolete. Applications should use [getaddrinfo(3)](http://man7.org/linux/man-pages/man3/getaddrinfo.3.html), [getnameinfo(3)](http://man7.org/linux/man-pages/man3/getnameinfo.3.html), and [gai_strerror(3)](http://man7.org/linux/man-pages/man3/gai_strerror.3.html) instead.

The changes merged in #87 introduce the use of `gethostbyname()` in [tcp_client.cpp](https://github.com/ros-industrial/industrial_core/blob/079a7ccc4fac2a85feed30e965c4cd1c2f29798e/simple_message/src/socket/tcp_client.cpp#L80). This should be fixed.

PS: [udp_client.cpp](https://github.com/ros-industrial/industrial_core/blob/12a74a1f9f26aea0ee075edaf4c84473bd8e112a/simple_message/src/socket/udp_client.cpp#L75) does not use `gethostbyname()`, but it seems like it should (well, it shouldn't, but it currently does not support using hostnames, like `tcp_client` does)?

machinekoder:
I start to work on this for #wrid18 now.

gavanderhoorn:
Fixed by #197.

PR #197, opened 2018-07-11
simple_message: get rid of obsolete gethostbyname

machinekoder:
fix for https://github.com/ros-industrial/industrial_core/issues/108

NOTE: For some reason the run_tests fail on my system, but they also fail without the patch, so I want to see how the CI handles this patch.

### DIFF
--- a/simple_message/include/simple_message/socket/simple_socket.h
+++ b/simple_message/include/simple_message/socket/simple_socket.h
@@ -53,8 +53,8 @@
 
 #define SOCKET(domain, type, protocol) socket(domain, type, protocol)
 #define BIND(sockfd, addr, addrlen) bind(sockfd, addr, addrlen)
-#define SET_NO_DELAY(sockfd, val) setsockopt(sockfd, IPPROTO_TCP, TCP_NODELAY, &val, sizeof(val))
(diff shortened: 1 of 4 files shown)


gavanderhoorn:
Thanks for the PR, and thanks for your WRID18 participation.

Would it be nice to print an info/warning message whenever name resolution doesn't work?

> NOTE: For some reason the run_tests fail on my system, but they also fail without the patch, so I want to see how the CI handles this patch.

seems CI is happy with your changes, I see a checkmark.

machinekoder:
@gavanderhoorn The current behavior implicitly assumes that the `buff` content is either a hostname which can be resolved or an IP address. So printing a warning would cause a warning to be printed whenever one wants to use an explicit IP-address.

So either we change the code to reverse the behavior, first trying to use the buffer contents as an IP-address, then trying to resolve it or it would change the current behavior.

gavanderhoorn (approved):
Works as described.

Printing warnings or errors when hosts cannot be resolved (or upon any other error) can be added in a later PR.

gavanderhoorn:
Travis failed because of a time-out, not because of something in this PR.

I'll merge.

gavanderhoorn:
Thanks again @machinekoder :+1:

Issue #175, opened 2017-03-29
robot_client: JointTrajectoryStreamer does not send TRAJ_STOP if in IDLE state


gavanderhoorn:
As per subject really.

If `JointTrajectoryStreamer::jointTrajectoryCB(..)` receives an empty `JointTrajectory` msg while the streamer state machine is in the `IDLE` state, it does not call `trajectoryStop()`, resulting in no `STOP_TRAJECTORY` message to the trajectory server (see [here](https://github.com/ros-industrial/industrial_core/blob/d18b676f30616bb29b92dede04120b364907f8c7/industrial_robot_client/src/joint_trajectory_streamer.cpp#L74-L85)). Sending an empty trajectory is used by `JointTrajectoryAction::cancelCB(..)` to stop execution of the currently executing trajectory ([here](https://github.com/ros-industrial/industrial_core/blob/aca48389468a53ca932373b8f94f0b3ea1bdc9db/industrial_robot_client/src/joint_trajectory_action.cpp#L183-L201)).

For drivers with points buffers on the controller the streamer could already be in `IDLE` while the robot is still executing the trajectory (as the last `N` points of the traj have been buffered). This leads to the `STOP` never being transmitted and thus the cancellation of goals potentially being ignored.

gavanderhoorn:
The question is thus: is this desired behaviour? Can any of the authors (@shaun-edwards, @JeremyZoss) perhaps comment on what the rationale was to ignore empty trajectories in `IDLE`?

Should empty messages not be used to cancel goals?

Or was the "point buffer on server" case not considered when this code was written?

JeremyZoss:
+1

I think you are correct, that this case (buffered points on controller) was not considered.  It's a bit difficult to review the history, since Google Code is mostly shutdown.  But it looks like `industrial_robot_client` mostly replicated the existing behavior of the original motoman `joint_trajectory_streamer` code.

I think an easy patch would be to:
  - revise the log message (line [89](https://github.com/ros-industrial/industrial_core/blob/d18b676f30616bb29b92dede04120b364907f8c7/industrial_robot_client/src/joint_trajectory_streamer.cpp#L89))
 - add a `trajectory_stop` block just below that message, similar to what's currently at lines [81-83](https://github.com/ros-industrial/industrial_core/blob/d18b676f30616bb29b92dede04120b364907f8c7/industrial_robot_client/src/joint_trajectory_streamer.cpp#L81-L83)

[...]

PR #199, opened 2018-07-11
Always call trajectoryStop if an empty joint traj message is received

schornakj:
Fix for issue #175. It looks like the desired behavior is:
- If the trajectory message is empty, we should stop the current trajectory regardless of the current state.
- If the message is not empty but the state is something other than IDLE, we should log an error to tell the user that trajectory splicing isn't currently implemented, and also stop the current trajectory.

### DIFF
--- a/industrial_robot_client/src/joint_trajectory_streamer.cpp
+++ b/industrial_robot_client/src/joint_trajectory_streamer.cpp
@@ -71,23 +71,25 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
   int state = this->state_;
 
   ROS_DEBUG("Current state is: %d", state);
-  if (TransferStates::IDLE != state)
-  {
-    if (msg->points.empty())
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Thanks for the PR.

I noticed that your commit isn't actually attributed to your account (uses your `@wpi` address). Would you want to fix that first?

schornakj:
Thank you for checking account attribution - I'll fix that later today and update.

schornakj:
@gavanderhoorn Just tying up loose ends -- do these changes still look OK?

Levi-Armstrong:
@gavanderhoorn Any objections to merging this?

schornakj:
@gavanderhoorn I was going through my open PRs and I noticed that this one is still unresolved. Looks like it just needs someone authoritative to review it. Is this still a fix we want to implement?

gavanderhoorn:
Rebased version in #260 (with some additional commits).

gavanderhoorn:
Thanks for the initial PR @schornakj :+1:

PR #206, opened 2018-08-09
fix CMakeLists files: unconfigured build_depends

ipa-nhg:


industrial_robot_simulator/CMakeLists.txt
```diff
@@ -2,11 +2,11 @@ cmake_minimum_required(VERSION 2.8.12)
 
 project(industrial_robot_simulator)
 
-find_package(catkin REQUIRED COMPONENTS std_msgs sensor_msgs
-  control_msgs trajectory_msgs)
+find_package(catkin REQUIRED COMPONENTS std_msgs sensor_msgs 
+  control_msgs trajectory_msgs industrial_msgs)
```
gavanderhoorn, line 6: This is a pure Python package, so the best thing to do would probably be to remove all `COMPONENTS` from the `find_package(..)` call, instead of adding `industrial_msgs`.

Then make sure all msg packages are listed as `exec_depend` in the manifest.

industrial_robot_simulator/CMakeLists.txt
```diff
@@ -2,11 +2,11 @@ cmake_minimum_required(VERSION 2.8.12)
...
 
 catkin_package(
-    CATKIN_DEPENDS std_msgs sensor_msgs control_msgs trajectory_msgs
+    CATKIN_DEPENDS std_msgs sensor_msgs control_msgs trajectory_msgs industrial_msgs
```
gavanderhoorn, line 9: Same here.

gavanderhoorn (changes requested):

industrial_robot_simulator/CMakeLists.txt
ipa-nhg: Yes! you are right it a python package. 

Actually the package.xml file is also wrong, all the dependencies are declared as `depend` instead of as `exec_depend`.

I will open a new PR and close this

ipa-nhg:
This PR is wrong, issue solved by https://github.com/ros-industrial/industrial_core/pull/207

PR #208, opened 2018-08-09
add roslaunch testing to industrial_robot_client

ipa-nhg:


industrial_robot_client/test/roslaunch_test.xml
```diff
@@ -0,0 +1,25 @@
...
+      <arg name="robot_ip" value="$(arg robot_ip)" />
+    </include>
+  </group>
+
+  <group ns="robot_state_visualize_">
+    <include file="$(find industrial_robot_client)/launch/robot_state_visualize.launch">
+      <arg name="robot_ip" value="$(arg robot_ip)" />
+      <arg name="urdf_path" value="$(arg urdf_path)" />
+    </include>
+  </group>
```
gavanderhoorn: See my comment on #209 for why I believe we should not include this test here.

industrial_robot_client/package.xml
```diff
@@ -34,4 +34,5 @@
   <run_depend>urdf</run_depend>
   <run_depend>industrial_msgs</run_depend>
   <run_depend>industrial_utils</run_depend>
+  <test_depend>roslaunch</test_depend>
```
gavanderhoorn: I seem to remember that for package manifest v1, this should actually be a `build_depend`. See https://github.com/ros-industrial/fanuc/issues/76 fi.

Moving to format v2 would allow us to keep the `test_depend` here, but let's not mix things and use a `build_depend` here.

ipa-nhg:
@gavanderhoorn please merge first the PR https://github.com/ros-industrial/industrial_core/pull/209 that will create a new merge conflict...

ipa-nhg:
@gavanderhoorn this should be now right

gavanderhoorn:
Looks alright, yes. Let's wait on Travis.

gavanderhoorn (approved):

gavanderhoorn:
And thanks again @ipa-nhg :cake:

PR #209, opened 2018-08-09
Added missed dependencies

ipa-nhg:
see https://travis-ci.org/ros-industrial/industrial_core/jobs/414064619#L3473


industrial_robot_client/package.xml
```diff
@@ -34,4 +34,6 @@
   <run_depend>urdf</run_depend>
   <run_depend>industrial_msgs</run_depend>
   <run_depend>industrial_utils</run_depend>
+  <run_depend>rviz</run_depend>
```
gavanderhoorn: Something to consider: in light of ros-industrial/ros_industrial_issues#50, we might not want to add `rviz` here (as that would make RViz again a dependency of all robot drivers that depend on `industrial_robot_client`).

gavanderhoorn (changes requested):
I'd merge if `rviz` is removed.

I realise that will make #208 fail, but we should remove the `robot_state_visualize` launch file from the test.

gavanderhoorn:
It seems the merge of #205 introduced a merge conflict. Could I ask you to fix that @ipa-nhg?

gavanderhoorn (approved):

gavanderhoorn:
Thanks @ipa-nhg :+1:

PR #212, opened 2018-10-05
Added in better simluation when velocities are provided for trajectory

pbeeson:
The current ROS-I sim just linearly interpolates as a simulator.  When velocities are provided in addition to positions, a better approach can be used that more accurately reflects the real robot motion.

See Section 2.7.2 from here: http://ros.informatik.uni-freiburg.de/roswiki/attachments/motoman_driver/MotoRos_EDS.pdf

This pull request uses this motion model when velocities are present, and defaults to the linear interpolation when they are not.  This has been tested and works well.

### DIFF
--- a/industrial_robot_simulator/industrial_robot_simulator
+++ b/industrial_robot_simulator/industrial_robot_simulator
@@ -141,6 +141,20 @@ class MotionControllerSimulator():
 
     """
     """
+    def accelerate(self, last_pt, current_pt, current_time, delta_time):
+        intermediate_pt = JointTrajectoryPoint()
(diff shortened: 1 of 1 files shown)


pbeeson:
The motion model used assumes linear acceleration between waypoint.

shaun-edwards (approved):

shaun-edwards:
@pbeeson, thanks for the addition.  +1 from me.  @gavanderhoorn, do you want to have a look?  Otherwise, I'll merge in a couple of days.

gavanderhoorn:
Late, but :+1: .

Thanks @pbeeson.

PR #214, opened 2018-11-01
Decrease idle wait time.  250 ms is too slow

pbeeson:
This is addressing Issue #191


simonschmeisser:
Thanks for the PR, we changed that in our fork at https://github.com/isys-vision/industrial_core/commit/d4025b03b6f2bccd1be4321a645810549787936f but I forgot to create a PR here

works fine for us on multiple fanucs, +1

gavanderhoorn (approved):

gavanderhoorn:
Thanks for the PR @pbeeson.

The `250` ms was used there when back-to-back trajectories were not a use-case for that node.

PR #217, opened 2018-11-26
Trajectory filters: check output of update method

marip8:
I recently ran into an issue with a trajectory filter (uniform sampling filter) which had been accidentally given a bad input trajectory such that the `update` method returned false. The `adaptAndPlan` method did not check the output of the `update` method which subsequently resulted in an exception being thrown elsewhere in the `move_group` code. This PR checks the call to the `update` method before continuing to write the new output trajectory to avoid issues later in the planning/execution pipeline


gavanderhoorn (approved):
Makes sense (and should've already been checked).

gavanderhoorn:
Thanks @marip8 :+1:

PR #222, opened 2019-02-28
all: update maintainer email addresses.

gavanderhoorn:
As per subject.

Noticed that Shaun's old SwRI address was still listed in all packages.

### DIFF
--- a/industrial_core/package.xml
+++ b/industrial_core/package.xml
@@ -3,10 +3,10 @@
   <version>0.5.1</version>
   <description>ROS-Industrial core stack contains packages and libraries for supporing industrial systems</description>
 
-  <maintainer email="sedwards@swri.org">Shaun Edwards</maintainer>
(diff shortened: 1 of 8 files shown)


robinsonmm:
Thanks, and we will have the team take a closer look moving forward. Shaun
has expressed interest in continuing to maintain Motoman related repos, but
even those should be updated to his current email address.

MR

gavanderhoorn:
Can we please get this merged first? I'd like to not have invalid email addresses in package manifests, even if the listed address is not the preferred maintainer any longer.

We can then start the discussion on who should be listed where.

gavanderhoorn:
To move forward, I need a review from at least one other maintainer with write access.

gavanderhoorn:
:bellhop_bell: @Levi-Armstrong or @shaun-edwards.

jrgnicho (approved):

jrgnicho:
@gavanderhoorn thanks for the update

PR #223, opened 2019-03-26
msgs: remove unneeded genmsg dependency.

gavanderhoorn:
As per subject.


jrgnicho (approved):

jrgnicho:
Here is an explanation about [genmsg and message_generation](https://answers.ros.org/question/65236/what-is-difference-between-message_generation-and-genmsg/)

gavanderhoorn:
Thanks @jrgnicho.

re: explanation: yes, I sort of assumed that would be clear, but should've probably added (or at least referenced) something.

PR #226, opened 2019-05-06
fix #225: incorrect return value if deserializing duration fails

### LINKED ISSUE #225: Incorrect return value on deserializing JointTrajPt
I am not entirely sure, but this line looks like a bug to me:

https://github.com/ros-industrial/industrial_core/blob/a588028a74a654af3dc06803b9cf6031d3b960e8/simple_message/src/joint_traj_pt.cpp#L111

It will return `true` even if the duration failed to load on the buffer. I don't know if this is for some reason an intended behavior, but it does feel more like a copy&paste bug. In that case, the fix should probably be to flatten this very nested construct to avoid this kind of bugs from creeping in.

gonzalocasas:
As discussed in #225, the `JointTrajPt` will return `true` even if deserializing `duration` failed.
This is the smallest possible fix, just removing the line that incorrectly set `rtn=true`.

### DIFF
--- a/simple_message/src/joint_traj_pt.cpp
+++ b/simple_message/src/joint_traj_pt.cpp
@@ -108,7 +108,6 @@ bool JointTrajPt::load(industrial::byte_array::ByteArray *buffer)
           rtn = false;
           LOG_ERROR("Failed to load joint traj pt. duration");
         }
-        rtn = true;
       }
       else
       {


gavanderhoorn:
Travis problems .. I've restarted it (known problem).

gavanderhoorn:
`rosdep`/GH is being annoying.

---

Edit: it's actually GH probably.

gonzalocasas:
@gavanderhoorn ping 😉

gavanderhoorn:
Restarted Travis. Let's see whether things will compile this time.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @gonzalocasas for the fix.

Much appreciated :+1:

PR #229, opened 2019-05-24
fix #228: correct off-by-one on debug output

### LINKED ISSUE #228: Incorrect logging output
The `joint_trajectory_streamer` outputs an off-by-one debugging line when sending points to the controller, where it says `"Point[%d of %d] sent to controller"` and it uses 0-based indexing for the first value, but `size` for the second value, e.g. a trajectory of 3 points will print the final point as  `"Point[2 of 3] sent to controller"`, which is misleading.

The bug is [on this line](https://github.com/ros-industrial/industrial_core/blob/kinetic-devel/industrial_robot_client/src/joint_trajectory_streamer.cpp#L194), instead of the index of the current point, it should output index+1, or move the increment line above the output.

gonzalocasas:
Trivial fix for debugging output.

### DIFF
--- a/industrial_robot_client/src/joint_trajectory_streamer.cpp
+++ b/industrial_robot_client/src/joint_trajectory_streamer.cpp
@@ -191,9 +191,9 @@ void JointTrajectoryStreamer::streamingThread()
         ROS_DEBUG("Sending joint trajectory point");
         if (this->connection_->sendAndReceiveMsg(msg, reply, false))
         {
+          this->current_point_++;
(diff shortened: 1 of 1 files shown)


gavanderhoorn (approved):
Order of statements is slightly confusing on first read, but logically correct, so :+1:.

gavanderhoorn:
Guess we don't look at debug output often enough to notice this.

gavanderhoorn:
And thanks for this one as well @gonzalocasas :+1:

Issue #238, opened 2019-08-21
TcpClient won't re-establish connection if physical connection is unplugged, then restored


PGlaubitzWork:
Observed running against ROS indigo. Ethernet cable between PC and Fanuc controller became disconnected. isConnected() properly reported loss of connection. Upon re-connection makeConnect() could not re-establish connection since socket handle was for lost connection.

PGlaubitzWork:
Proposed fix in PR #239

haudren:
Sorry to do some archaeology here, but I just encountered a similar scenario when running `fanuc_driver` against a robot in an unstable network setting. I would sometimes get the following warnings and errors:
```
1599809178.461369439 WARN /joint_state [industrial_core/simple_message/src/socket/simple_socket.cpp:140(simple_socket::SimpleSocket::receiveBytes)] [topics: /rosout, /feedback_states, /joint_states, /robot_status] Recieved zero bytes: 0
1599809178.461382919 ERROR /joint_state [industrial_core/simple_message/src/smpl_msg_connection.cpp:118(smpl_msg_connection::SmplMsgConnection::receiveMsg)] [topics: /rosout, /feedback_states, /joint_states, /robot_status] Failed to receive message length
1599809178.461388687 ERROR /joint_state [industrial_core/simple_message/src/message_manager.cpp:166(message_manager::MessageManager::spinOnce)] [topics: /rosout, /feedback_states, /joint_states, /robot_status] Failed to receive incoming message
1599809178.461393753 WARN /joint_state [industrial_core/simple_message/include/simple_message/simple_comms_fault_handler.h:87(simple_comms_fault_handler::SimpleCommsFaultHandler::sendFailCB)] [topics: /rosout, /feedback_states, /joint_states, /robot_status] Send failure, no callback support
```
That would then be followed by a flurry of:
```
1599209056.621394184 ERROR /joint_state [include/simple_message/socket/simple_socket.h:256(simple_socket::SimpleSocket::logSocketError)] [topics: /rosout, /feedback_states, /joint_states, /robot_status] Failed to connect to server, rc: -1. Error: 'Transport endpoint is already connected' (errno: 106)
```
I tried to enable REUSE_ADDR in the client, but that only resulted in the following:
```
[...]

PR #239, opened 2019-08-21
Issue-238 Fix tested  on indigo & Fanuc

PGlaubitzWork:
`connectSocketHandle()` will close and re-establish socket handle if the handle is from a previous connection, otherwise it will establish the socket handle.

For #238.

### DIFF
--- a/simple_message/include/simple_message/socket/tcp_client.h
+++ b/simple_message/include/simple_message/socket/tcp_client.h
@@ -76,6 +76,9 @@ public:
     // Overrides
     bool makeConnect();
 
+private:
+    bool connectSocketHandle();
+
 
 };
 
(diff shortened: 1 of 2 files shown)


haudren:
@gavanderhoorn @Levi-Armstrong : Could you maybe have a look at this PR? it does indeed fix the issue we have with the FANUC robot when the connection drops. I am at your disposal if you need help reproducing the issue and so on.

haudren:
@gavanderhoorn @Levi-Armstrong : Who should I contact to get this pushed forward? I don't mind taking care of any necessary changes if needs be.

JeremyZoss:
@Levi-Armstrong I don't currently have a good system configured to run tests on this.  But a read-through of the changes looks okay to me.  A little caution is prudent, since this code is used by many drivers.  But socket disconnect handling has long been a weak point of the drivers, so improvements are welcomed.  Was independently verified in 2 different systems (both Fanuc) - see comments on #238.  My vote is we merge this as-is and keep an eye out for new corner cases that may have been introduced.

@PGlaubitzWork Thanks for this submission!

gavanderhoorn:
Replacement: #263.

Thanks for the initial PR @PGlaubitzWork :+1:

Issue #34, opened 2013-08-23
simple_message should include receive timeout


JeremyZoss:
Currently, the simple_message receive calls will block forever until the requested message data is received or the connection is dropped/reset.  For some applications, it may be helpful for the user to be able to specify a timeout on the socket-read.

The recent fix to Issue #25 helps provide some of the low-level socket structure needed for this enhancement.  But, this enhancement will require additional changes to the higher-level messaging layers to pass the timeout error status up to user-level code.

ghost:
I would also like this feature!

PR #246, opened 2019-11-21
Adds receive overloads with timeouts

ghost:
See also #34. 

Existing receive functions are unaffected, and behave exactly as before. The only big difference is that all classes inheriting from `SmplMsgConnection`, i.e., `SimpleSocket` must now
override a `receiveBytes` member function with a timeout parameter. (So if any other classes inherit from `SmplMsgConnection`, they will need to be updated.)

Also, a few random whitespace fixes I couldn't resist.

### DIFF
--- a/simple_message/include/simple_message/smpl_msg_connection.h
+++ b/simple_message/include/simple_message/smpl_msg_connection.h
@@ -64,7 +64,7 @@ class SmplMsgConnection
 public:
 
   // Message
-  
+
   /**
    * \brief Sends a message using the data connection
    *
@@ -73,7 +73,7 @@ public:
    * \return true if successful
    */
   virtual bool sendMsg(industrial::simple_message::SimpleMessage & message);
-  
+
   /**
    * \brief Receives a message using the data connection
    *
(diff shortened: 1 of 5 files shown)


ghost:
Ping!

ghost:
If anyone is wondering why this is necessary, consider the following code:

```
if (conn->isReadyReceive(ms_timeout)) {
  conn->receiveMsg(res);
}
```

If you were hoping this code would always result in a non-blocking read, well that's not guaranteed. 

The problem is that `isReadyReceive` can only check if *any* data is ready for reading.
On the other hand, `receiveMsg` will block until *all* data has been read.

So if you receive the first few bytes of a message, but then the connection hangs:

1. `isReadyReceive` will call `rawPoll` and return true.
2. `receiveMsg` will call `receiveBytes`, which will enter its internal while loop.
3. `receiveBytes` will successfully receive a few bytes.
4. (Connection hangs here.)
5. `receiveBytes` will no longer getting any bytes.
6. Stuck forever in `while (remainBytes > 0) {`.

The problem is that there is no way to escape the internal while loop, and doing `isReadyReceive` doesn't actually prevent `receiveMsg` from getting stuck. The problem is worse for large messages.

realanton (dismissed):
This is awesome- lets try to get this in asap and Ill rebase the yaskawa code I have on this. Thanks @tim-rtr

ghost:
@realanton I would if I could but this is actually the ROS Industrial repo! :laughing:

gavanderhoorn:
Newer version in #267.

PR #254, opened 2020-03-27
industrial_robot_simulator: listen to 'fake_e_stop' and change availability accordingly

simonschmeisser:
(cherry picked from commit 54740094729b9340000d2cc42f5a01f170392966)

This adds the ability to trigger and resolve an emergency stop in industrial_robot_simulator to test error handling offline

### DIFF
--- a/industrial_robot_simulator/industrial_robot_simulator
+++ b/industrial_robot_simulator/industrial_robot_simulator
@@ -49,6 +49,7 @@ from industrial_msgs.msg import RobotStatus
 # Subscribe
 from trajectory_msgs.msg import JointTrajectory
 from trajectory_msgs.msg import JointTrajectoryPoint
+from std_msgs.msg import Bool
 
 # Services
 from industrial_msgs.srv import GetRobotInfo, GetRobotInfoResponse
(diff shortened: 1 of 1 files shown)


industrial_robot_simulator/industrial_robot_simulator
```diff
@@ -270,7 +271,11 @@ class IndustrialRobotSimulatorNode():
         # retrieve update rate
         motion_update_rate = rospy.get_param('motion_update_rate', 100.);  #set param to 0 to ignore interpolated motion
         self.motion_ctrl = MotionControllerSimulator(num_joints, initial_joint_state, update_rate=motion_update_rate)
-
+        
+        # React to published e-stop
+        self.e_stop_state = False
+        self.e_stop_sub = rospy.Subscriber('fake_e_stop', Bool, self.e_stop_changed_callback)
```
gavanderhoorn, line 277: Would it be better to *not* call this `fake_e_stop`? The whole script is already a simulator/simulation of a real robot, and those typically don't have `fake_e_stop` topics.

If we want to add this here, I would propose the rename the subscribed topic. If the producing side would like to call the topic `fake_e_stop`, remapping could be used.

gavanderhoorn (changes requested):
See inline comment.

(the bigger discussion is that `industrial_robot_simulator` is supposed to implement a "standard interface", and e-stop topics are not part of whatever standard we're currently implementing, but that's perhaps out-of-scope for this PR)

PR #259, opened 2021-03-15
Remove exec permission on source files

galou:
Signed-off-by: Gaël Écorchard <gael.ecorchard@cvut.cz>


gavanderhoorn (approved):

gavanderhoorn:
Thanks, those files should indeed not be executable.

PR #260, opened 2021-06-18
Rebased 199: always send trajectory stop request

gavanderhoorn:
Regardless of the state in which the trajectory streamer node is (ie: idle or not).

This is a rebase of #199 by @schornakj with some extra (cosmetic) commits.


gavanderhoorn:
As this is my own PR now, I can't approve it.

I have however tested #199 on real hw (with `fanuc_driver`) and it appears to address the original issue (#175).

Based on that, I'm merging this one.

PR #262, opened 2021-06-22
Rebased 190: export compiler flags

gavanderhoorn:
This is a rebase of #190 (which addresses #188), combined with a migration to a more scalable solution (which itself is a compromise between how we'd like to do it (with modern CMake) and what Catkin supports in ROS 1).

I opted to include additional commits in this rebased PR as the basic principle (use a `CFG_EXTRAS` file to export `-D` flags to dependents) is still used. The file just doesn't set those flags directly, that's left up to the dependent.

I've also opted to remove the old defines (ie: `ROS`, `MOTOPLUS` and `LINUXSOCKETS`) instead of keeping them around.

They were really only used with early versions of MotoROS and we don't want to have them lingering around.

### DIFF
--- a/industrial_robot_client/CMakeLists.txt
+++ b/industrial_robot_client/CMakeLists.txt
@@ -66,9 +66,13 @@ install(TARGETS ${PROJECT_NAME}_dummy DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATI
 
 add_library(industrial_robot_client ${SRC_FILES})
 target_link_libraries(industrial_robot_client simple_message)
+target_compile_definitions(industrial_robot_client PUBLIC
+  ${simple_message_DEFINITIONS})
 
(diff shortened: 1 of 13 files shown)


gavanderhoorn:
Note: this does require dependents to actively use the exported `simple_message_DEFINITIONS` and `industrial_robot_client_DEFINITIONS` variables. But that seems like an OK compromise considering the alternative would be to unconditionally add the flags to the dependent's compilation flags.

I'll prepare a PR against `fanuc_driver` and `abb_driver` to update their build scripts.

gavanderhoorn:
@JeremyZoss @Levi-Armstrong: if either of you could review?

As this is now a PR I opened I cannot review it myself.

Levi-Armstrong (approved):

gavanderhoorn:
Thanks @Levi-Armstrong :+1:

PR #267, opened 2021-06-23
Rebased 246: receive timeouts for simple socket

### LINKED ISSUE #34: simple_message should include receive timeout

gavanderhoorn:
As per subject.

Commit provenance and attribution retained, although the original account appears to have been deleted.

One additional commit fixes some style issues.

### DIFF
--- a/simple_message/include/simple_message/smpl_msg_connection.h
+++ b/simple_message/include/simple_message/smpl_msg_connection.h
@@ -87,12 +87,12 @@ public:
    * \brief Receives a message using the data connection with a timeout.
    *
    * \param [out] message Populated with received message
-   * \param [in] timeoutMs The timeout for receiving a message, in milliseconds
+   * \param [in] timeout_ms The timeout for receiving a message, in milliseconds
    *
    * \return true if successful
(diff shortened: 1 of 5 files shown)


gavanderhoorn:
We'll merge this, although this isn't actually used anywhere. Not in IRC nor in any of the drivers we maintain which use `simple_message` or the IRC.

As such, it's difficult for me to test it, and I can only rely on the review that was already done in #246 and the unit test that was added as part of the PR (which is great of course).

gavanderhoorn:
As the original PR already had been reviewed, I'm going to merge this without an additional review.

gavanderhoorn:
Thanks Tim Perkins for the initial PR.

PR #269, opened 2021-06-23
ci: use catkin_tools for focal builds

gavanderhoorn:
As per subject.

Colcon seems to have trouble discovering all the tests.


gavanderhoorn:
216 tests now both on Bionic and Focal.

Merging.

PR #270, opened 2021-06-24
Enable catkin_lint

gavanderhoorn:
As per subject.

The `ignore_once` pragma in the `.launch` file requires an unreleased version of `catkin_lint` for now (fkie/catkin_lint#98), so this PR will fail CI.

We'll wait until the updated version is released.


gavanderhoorn:
This didn't fail because I didn't enable *strict* mode.

gavanderhoorn:
New version of `catkin_lint` -> tests pass.

Issue #265, opened 2021-06-23
trajectory relays may drop points after/while reconnecting


gavanderhoorn:
While reviewing #263, there have been a few cases where the relay dropped points it claimed to "retry at a later time".

From https://github.com/ros-industrial/industrial_core/pull/263#issue-675682823:

> the streaming trajectory relay actually drops traj pts when it loses connection. It claims to then "retry later", but it actually skips the point it dropped and continues with the next one in the trajectory. Before this change, that was not too much of a problem, as the connection would not come up again (requiring a restart, emptying queues and restarting everything).

this should be checked and fixed.

Issue #118, opened 2015-07-10
robot_client: joint_trajectory_streamer doesn't check reply msg from ctrlr in streamingThread


gavanderhoorn:
The state machine in `JointTrajectoryStreamer::streamingThread()` uses `SmplMsgConnection::sendAndReceiveMsg(req, reply, ..)` to send `JointTrajPtMessage` instances to the server program running on the controller, but the returned `reply` msg is never checked for success of the operation ([here](https://github.com/ros-industrial/industrial_core/blob/d18b676f30616bb29b92dede04120b364907f8c7/industrial_robot_client/src/joint_trajectory_streamer.cpp#L188-L199)). As trajectory streaming is done using (synchronous) `SERVICE_REQUEST` type msgs, it would seem logical to also check the reply from the controller.

The current implementation makes it impossible for servers to reject or _nack_ an incoming traj pt (or in general: a service request).

gavanderhoorn:
Do we actually care about this? Should the act of transferring the traj pt be considered the 'service request', and not so much the execution of the associated motion? If that is the case, the fields in the `RobotStatus` msgs would indirectly give feedback on the success of the motion (no errors on the `robot_status` topic == success), and there is no need to be able to nack points explicitly.

@shaun-edwards: could you elaborate on the design of the protocol wrt this?

shaun-edwards:
The success should by tied to the queuing of the motion, since this is often times the only information the controller provides.  Most controllers queue up motion separate from program execution.  As a result, sending a response, when the motion command is executed/achieved is no easily done.  Should anything fail in the execution of a trajectory, the `RobotStatus` should indicate this.

gavanderhoorn:
So it wouldn't make much sense to check at service invocation if a traj pt is actually reachable by the manipulator, servers should just queue it and fail whenever the motion is actually attempted?

Depending on buffer size that could mean traj execution fails after quite some time.

Would it perhaps make sense to rename the msgs then? Instead of `JOINT_TRAJ_PT` something like `QUEUE_X` (where `X` is a `TRAJ_PT`, `DYN_JNT_PT` etc) would seem to be much clearer as it conveys what is actually the idea behind the service request.

gavanderhoorn:
This is slightly off-topic for this issue actually, as even if only the queueing itself should be considered the service req, the current client nodes still do not check the response.

Issue #131, opened 2016-01-18
JointTrajectoryDownloader/Streamer provides no feedback


simonschmeisser:
If there is an error during upload of the trajectory, there is no way for JointTrajectoryDownloader to communicate this error. This happens for me ie when the controller is set to "Manual mode" instead of "automatic". The mitsubishi controller will allow for upload but returns an error once I try to start execution of the trajectory. This does not lead to the controller being in an error state, which could be detected by RobotStateInterface and handled by pr #83. So these two are complementary.

I would personally implement this using (1) a service call instead of a one way topic, but that would probably break compatibility in drivers. So instead (2) a separate return topic could be used with the Downloader/Streamer publishing errors. JointTrajectoryAction would then listen for these and react in a way similar to pr #83.

gavanderhoorn:
Just providing my insight here:

> If there is an error during upload of the trajectory, there is no way for JointTrajectoryDownloader to communicate this error.

Well, with msgs like `JointTrajPt(Full)` being a `SERVICE_REQUEST`, the idea is that server implementations `ACK` or `NACK` the action of _enqueuing points_. In your situation (upload ok, exec not) that doesn't help, but it does allow the server to report either success or failure during the streaming/uploading process. Support for this is currently incomplete in the generic clients (#118), but the protocol allows for it.

> This happens for me ie when the controller is set to "Manual mode" instead of "automatic". The mitsubishi controller will allow for upload but returns an error once I try to start execution of the trajectory. This does not lead to the controller being in an error state, which could be detected by RobotStateInterface and handled by pr #83. So these two are complementary.

`STATUS` includes a field `motion_possible`, which is supposed to be set to `TRUE` only if the controller is able to execute trajectories (ie: move). If for Mitsubishi controllers being able to move includes things like not being in manual mode, then it should be `FALSE` in those cases.

I've drivers for controllers where being in manual mode is fine, as long as the operator is pressing some keys on the teach pendant. The condition (in the driver) for setting `motion_possible` to `TRUE` includes the state of those keys, hence it will only be `TRUE` if motion is actually possible. In addition: attempting motion execution while the operator is not holding those keys on some controllers results in an error, which should then also be reported using the `in_error` and `error_code` fields.

Again, support for this in the generic clients is incomplete, and #83 rectifies some of that.

Nothing currently checks the status of the `motion_possible` field though, which I think is crucial to remedying your issue.

[...]

PR #271, opened 2021-06-26
Sort-of rebase of 83: traj action aborts on error

gavanderhoorn:
Sort-of, as it's actually a bit more than what was included in #83.

The proposed changes make the `joint_trajectory_action` aware of the state of the (OEM) motion server program (as far as it publishes that in the form of `RobotStatus` messages). In two cases is this information taken into account:

 1. while accepting new goals
 2. while executing already accepted goals

In both cases, the following is monitored / checked:

 1. is an e-stop active?
 2. is there some other error active?
 3. does the motion server report `motion_possible` is `true`?

If any of these are not as expected, the goal is rejected, or, if a goal is currently being processed (ie: motion is execution), the goal is aborted and trajectory execution cancelled (as that should now work, with #260 merged (which still requires a cooperating driver of course)).

In all cases, the action result would be `INVALID_GOAL`, as unfortunately the `FollowJointTrajectory` action message does not define any really more suitable error codes.

Also in all cases: the reason for aborting or rejecting a goal is described in a human readable message.

When rejecting goals:

```
[ERROR] [...]: Rejecting goal: controller reported motion not possible (no further information)
[ERROR] [...]: Rejecting goal: controller reported (active) error (OEM code: 11003)
[ERROR] [...]: Rejecting goal: controller reported e-stop
```

When aborting goals:

```
[ERROR] [...]: Aborting goal: controller reported motion not possible (no further information)
[ERROR] [...]: Aborting goal: controller reported (active) error (OEM code: 11003)
[ERROR] [...]: Aborting goal: controller reported e-stop
```

(this is output from `fanuc_driver`: `11003` is *Deadman switch released*, which will indeed prevent any motion in manual mode)

While this is certainly not perfect (it doesn't address #265 nor #118), together with #263 and #260 and an (OEM) relay which behaves according to the spec, the UX of `industrial_robot_client` is improved quite a bit.

Lastly: as previous versions of the JTA did not check anything, misbehaving drivers would still be able to execute motions, even if they didn't properly populate all fields of the `RobotStatus` messages. In order to still allow usage of such systems, two new parameters were added which can be used to revert the JTA back to its previous behaviour (ie: don't check anything, always forward goals to drivers).

In the current implementation, the default is to disable these new checks. Something to discuss is whether we should make the default to *enable* the new checks, and users with misbehaving drivers should instead be required to explicitly enable the bw-compatible behaviour.

### DIFF
--- a/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
@@ -188,7 +188,7 @@ private:
    * cannot accurately report controller status (ie: have UNKNOWNs in
    * the RobotStatus messages they publish).
    */
-  bool ignore_motion_server_error_;
+  bool ignore_motion_server_error_ = false;
 
   /**
    * \brief Should the action server consider UNKNOWN for TriStates in
@@ -201,7 +201,7 @@ private:
    * are unable to accurately report controller status, by allowing
    * the action server to assume UNKNOWN == OK.
    */
-  bool consider_status_unknowns_ok_;
+  bool consider_status_unknowns_ok_ = false;
 
   /**
    * \brief Watch dog callback, used to detect robot driver failures
--- a/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
@@ -246,6 +246,11 @@ private:
   void robotStatusCB(const industrial_msgs::RobotStatusConstPtr &msg);
 
   /**
+   * \brief Sends a stop command (empty message) to the robot driver.
+   */
+  void stopRelay();
+
+  /**
    * \brief Aborts the current action goal and sends a stop command
    * (empty message) to the robot driver.
    *
--- a/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
@@ -182,7 +182,8 @@ private:
    *
    * This is a backwards compatibility 'tunable knob'.
    *
-   * Set this to true to keep the old behaviour.
+   * The constructor overrides the default here with the value of the private
+   * parameter on the parameter server with the same name.
    *
    * This is configurable, as it's possible drivers for certain robots
    * cannot accurately report controller status (ie: have UNKNOWNs in
@@ -195,6 +196,9 @@ private:
    * a RobotStatus message as OK?
    *
    * This is a backwards compatibility 'tunable knob'.
+   *
+   * The constructor overrides the default here with the value of the private
+   * parameter on the parameter server with the same name.
    *
    * Use this to effectively ignore UNKNOWN states for TriStates in
    * RobotStatus messages. This can help with OEM server programs which
--- a/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
+++ b/industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
@@ -177,6 +177,33 @@ private:
   static const double WATCHDOG_PERIOD_;// = 1.0;
 
   /**
+   * \brief Should the action server abort goals if the OEM server
+   * program reports a problem?
+   *
+   * This is a backwards compatibility 'tunable knob'.
+   *
+   * Set this to true to keep the old behaviour.
+   *
+   * This is configurable, as it's possible drivers for certain robots
+   * cannot accurately report controller status (ie: have UNKNOWNs in
+   * the RobotStatus messages they publish).
+   */
+  bool ignore_motion_server_error_;
+
+  /**
+   * \brief Should the action server consider UNKNOWN for TriStates in
+   * a RobotStatus message as OK?
+   *
+   * This is a backwards compatibility 'tunable knob'.
+   *
+   * Use this to effectively ignore UNKNOWN states for TriStates in
+   * RobotStatus messages. This can help with OEM server programs which
+   * are unable to accurately report controller status, by allowing
+   * the action server to assume UNKNOWN == OK.
+   */
+  bool consider_status_unknowns_ok_;
+
+  /**
    * \brief Watch dog callback, used to detect robot driver failures
    *
    * \param e time event information
--- a/industrial_robot_client/include/industrial_robot_client/utils.h
+++ b/industrial_robot_client/include/industrial_robot_client/utils.h
@@ -36,7 +36,7 @@
 #include <string>
 #include <map>
 
-#include <industrial_msgs/RobotStatus.h>
+#include <industrial_msgs/TriState.h>
 
 
 namespace industrial_robot_client
--- a/industrial_robot_client/include/industrial_robot_client/utils.h
+++ b/industrial_robot_client/include/industrial_robot_client/utils.h
@@ -126,7 +126,7 @@ bool isWithinRange(const std::vector<std::string> & lhs_keys, const std::vector<
  *
  * \return true if the state is UNKNOWN
  */
-bool is_unknown(industrial_msgs::TriState const& state)
+bool isUnknown(industrial_msgs::TriState const& state)
 {
   return state.val == industrial_msgs::TriState::UNKNOWN;
 }
@@ -139,10 +139,10 @@ bool is_unknown(industrial_msgs::TriState const& state)
  *
  * \return true if the state is ON
  */
-bool is_on(industrial_msgs::TriState const& state, bool unknown_is_on)
+bool isOn(industrial_msgs::TriState const& state, bool unknown_is_on)
 {
   return state.val == industrial_msgs::TriState::ON
-    || (unknown_is_on && is_unknown(state));
+    || (unknown_is_on && isUnknown(state));
 }
 
 /**
@@ -153,10 +153,10 @@ bool is_on(industrial_msgs::TriState const& state, bool unknown_is_on)
  *
  * \return true if the state is OFF
  */
-bool is_off(industrial_msgs::TriState const& state, bool unknown_is_off)
+bool isOff(industrial_msgs::TriState const& state, bool unknown_is_off)
 {
   return state.val == industrial_msgs::TriState::OFF
-    || (unknown_is_off && is_unknown(state));
+    || (unknown_is_off && isUnknown(state));
 }
 
 } //utils
--- a/industrial_robot_client/include/industrial_robot_client/utils.h
+++ b/industrial_robot_client/include/industrial_robot_client/utils.h
@@ -119,6 +119,9 @@ bool isWithinRange(const std::vector<std::string> & lhs_keys, const std::vector<
                    const std::vector<std::string> & rhs_keys, const std::vector<double> & rhs_values,
                    double full_range);
 
+
+namespace tri_state {
+
 /**
  * \brief Check whether the given TriState is set to UNKNOWN.
  *
@@ -158,6 +161,8 @@ bool isOff(industrial_msgs::TriState const& state, bool unknown_is_off)
   return state.val == industrial_msgs::TriState::OFF
     || (unknown_is_off && isUnknown(state));
 }
+
+} //tri_state
 
 } //utils
 } //industrial_robot_client
--- a/industrial_robot_client/include/industrial_robot_client/utils.h
+++ b/industrial_robot_client/include/industrial_robot_client/utils.h
@@ -36,6 +36,9 @@
 #include <string>
 #include <map>
 
+#include <industrial_msgs/RobotStatus.h>
+
+
 namespace industrial_robot_client
 {
 namespace utils
@@ -115,6 +118,46 @@ bool isWithinRange(const std::vector<std::string> & keys, const std::map<std::st
 bool isWithinRange(const std::vector<std::string> & lhs_keys, const std::vector<double> & lhs_values,
                    const std::vector<std::string> & rhs_keys, const std::vector<double> & rhs_values,
                    double full_range);
+
+/**
+ * \brief Check whether the given TriState is set to UNKNOWN.
+ *
+ * \param[in] state to check
+ *
+ * \return true if the state is UNKNOWN
+ */
+bool is_unknown(industrial_msgs::TriState const& state)
+{
+  return state.val == industrial_msgs::TriState::UNKNOWN;
+}
+
+/**
+ * \brief Check whether the given TriState is set to ON (or HIGH or TRUE ..).
+ *
+ * \param[in] state the state to check
+ * \param[in] unknown_is_on should UNKNOWN be considered ON?
+ *
+ * \return true if the state is ON
+ */
+bool is_on(industrial_msgs::TriState const& state, bool unknown_is_on)
+{
+  return state.val == industrial_msgs::TriState::ON
+    || (unknown_is_on && is_unknown(state));
+}
+
+/**
+ * \brief Check whether the given TriState is set to OFF (or LOW or FALSE ..).
+ *
+ * \param[in] state the state to check
+ * \param[in] unknown_is_off should UNKNOWN be considered OFF?
+ *
+ * \return true if the state is OFF
+ */
+bool is_off(industrial_msgs::TriState const& state, bool unknown_is_off)
+{
+  return state.val == industrial_msgs::TriState::OFF
+    || (unknown_is_off && is_unknown(state));
+}
 
 } //utils
 } //industrial_robot_client
--- a/industrial_robot_client/launch/robot_interface_download.launch
+++ b/industrial_robot_client/launch/robot_interface_download.launch
@@ -30,6 +30,9 @@
   <node pkg="industrial_robot_client" type="motion_download_interface" name="motion_download_interface"/>
   
   <!-- joint_trajectory_action: provides actionlib interface for high-level robot control -->
-  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action"/>
+  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action">
+    <param name="ignore_motion_server_error" type="bool" value="true" />
+    <param name="consider_status_unknowns_ok" type="bool" value="true" />
+  </node>
 
 </launch>
--- a/industrial_robot_client/launch/robot_interface_download.launch
+++ b/industrial_robot_client/launch/robot_interface_download.launch
@@ -18,6 +18,37 @@
   <!-- robot_ip: IP-address of the robot's socket-messaging server -->
   <arg name="robot_ip" doc="IP of the controller" />
   
+  <!-- Prior to https://github.com/ros-industrial/industrial_core/pull/271, the
+       joint_trajectory_action server ignored the state of the OEM server program
+       and essentially open-loop forwarded incoming trajectories to the other
+       nodes in industrial_robot_client.
+
+       The new behaviour is to monitor the OEM server programs and reject
+       incoming goals and abort active ones when needed.
+
+       Set this argument to true to revert back to the old behaviour.
+       If this argument is set to false, the new behaviour will be used.
+  -->
+  <arg name="ignore_motion_server_error" default="false"
+    doc="Should OEM controller state be monitored while accepting goals and during execution?" />
+  
+  <!-- Some OEM server programs (state relays running on the OEM controller)
+       cannot (reliably) determine values for certain fields in RobotStatus
+       messages. Those fields should then be set to TriState::UNKNOWN by those
+       servers.
+
+       UNKNOWN is essentially treated like OFF (or FALSE) by the code which
+       monitors the OEM server program, so for those server programs which must
+       send UNKNOWN, the monitoring code will always detect errors and refuse to
+       execute any trajectories.
+
+       Set this parameter to true to avoid this with such OEM server programs,
+       at the cost of potentially accepting goals and attempting to execute them
+       even if the server program would not be able to actually do that.
+  -->
+  <arg name="consider_status_unknowns_ok" default="false"
+    doc="Should UNKNOWNs in RobotStatus messages be considered as OKs?" />
+  
   <!-- copy the specified IP address to the Parameter Server, for use by nodes below -->
   <param name="/robot_ip_address" type="str" value="$(arg robot_ip)"/>
   
@@ -31,8 +62,8 @@
   
   <!-- joint_trajectory_action: provides actionlib interface for high-level robot control -->
   <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action">
-    <param name="ignore_motion_server_error" type="bool" value="true" />
-    <param name="consider_status_unknowns_ok" type="bool" value="true" />
+    <param name="ignore_motion_server_error" type="bool" value="$(arg ignore_motion_server_error)" />
+    <param name="consider_status_unknowns_ok" type="bool" value="$(arg consider_status_unknowns_ok)" />
   </node>
 
 </launch>
--- a/industrial_robot_client/launch/robot_interface_streaming.launch
+++ b/industrial_robot_client/launch/robot_interface_streaming.launch
@@ -30,7 +30,9 @@
   <node pkg="industrial_robot_client" type="motion_streaming_interface" name="motion_streaming_interface"/>
   
   <!-- joint_trajectory_action: provides actionlib interface for high-level robot control -->
-  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action"/>
+  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action">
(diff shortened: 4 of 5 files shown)


gavanderhoorn:
Note: 55034f5 is basically what #83 proposed, but it also actually signals to the relay it should stop sending points.

Subsequent commits rewrite this into what the current implementation does.

gavanderhoorn:
I've requested @Levi-Armstrong and/or @JeremyZoss review this, as they've been involved with the development of `industrial_core` in the past.

@simonschmeisser: if you're still using this, and have some time, I'd appreciate your opinion as well.

Note: I do not intent to keep developing all of this. I just wanted to make sure I can get #83 in before releasing a new version of `industrial_core` into Melodic and Noetic. But since #83 actually only did half of what it was supposed to, I opted to flesh it out a bit more. And with #260 in place, it was actually not *that* difficult.

Note 2: it's still possible for trajectory relays to keep moving after an e-stop or error state is cleared: there is no way for the ROS side here (so the IRC node) to enforce anything. The streamer will now in all cases send a "traj abort" message, but if the trajectory and state relays (running on the OEM controller) do not cooperate, there isn't much we can do.

`fanuc_driver` fi will continue to execute the last sent point after resuming the relay program. Then it will stop. `fanuc_driver_exp` is slightly better, in that it purges its internal motion buffer and actually `SKIP`s the current segment. Other drivers may behave similarly (`abb_driver` fi).

gavanderhoorn:
This probably also addresses #131 partly, as this should not be possible any more (the JTA should reject the goal due to the controller not being in AUTO mode, or at least not reporting that `motion_possible` is `true`):

> This happens for me ie when the controller is set to "Manual mode" instead of "automatic". The mitsubishi controller will allow for upload but returns an error once I try to start execution of the trajectory. 

Of course it will depend on the (OEM) state program to properly report controller state (ie: `motion_possible` should accurately reflect whether motion is possible or not).

gavanderhoorn:
Friendly ping.

I don't believe this needs extensive testing.

The main question is whether we want to change the default behaviour (from not checking to checking).

industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
```diff
@@ -176,6 +176,33 @@ class JointTrajectoryAction
...
+   *
+   * This is a backwards compatibility 'tunable knob'.
+   *
+   * Set this to true to keep the old behaviour.
+   *
+   * This is configurable, as it's possible drivers for certain robots
+   * cannot accurately report controller status (ie: have UNKNOWNs in
+   * the RobotStatus messages they publish).
+   */
+  bool ignore_motion_server_error_;
```
simonschmeisser:
```suggestion
  bool ignore_motion_server_error_ = false;
```
I think it's already implicitly initialized to that but I'd prefer having this explicitely

industrial_robot_client/include/industrial_robot_client/joint_trajectory_action.h
```diff
@@ -176,6 +176,33 @@ class JointTrajectoryAction
...
+   * a RobotStatus message as OK?
+   *
+   * This is a backwards compatibility 'tunable knob'.
+   *
+   * Use this to effectively ignore UNKNOWN states for TriStates in
+   * RobotStatus messages. This can help with OEM server programs which
+   * are unable to accurately report controller status, by allowing
+   * the action server to assume UNKNOWN == OK.
+   */
+  bool consider_status_unknowns_ok_;
```
simonschmeisser:
```suggestion
  bool consider_status_unknowns_ok_ = true;
```

industrial_robot_client/launch/robot_interface_download.launch
```diff
@@ -30,6 +30,9 @@
   <node pkg="industrial_robot_client" type="motion_download_interface" name="motion_download_interface"/>
   
   <!-- joint_trajectory_action: provides actionlib interface for high-level robot control -->
-  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action"/>
+  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action">
+    <param name="ignore_motion_server_error" type="bool" value="true" />
```
simonschmeisser:
```suggestion
    <param name="ignore_motion_server_error" type="bool" value="false" />
```
I really appreciate this being tunable but I think we should be strict by default as it can introduce surprising delayed movements if you later enable movements (maybe  ... maybe not ... ?)

industrial_robot_client/launch/robot_interface_streaming.launch
```diff
@@ -30,7 +30,9 @@
   <node pkg="industrial_robot_client" type="motion_streaming_interface" name="motion_streaming_interface"/>
   
   <!-- joint_trajectory_action: provides actionlib interface for high-level robot control -->
-  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action"/>
+  <node pkg="industrial_robot_client" type="joint_trajectory_action" name="joint_trajectory_action">
+    <param name="ignore_motion_server_error" type="bool" value="true" />
```
simonschmeisser:
```suggestion
    <param name="ignore_motion_server_error" type="bool" value="false" />
```

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -45,12 +45,28 @@ const double JointTrajectoryAction::DEFAULT_GOAL_THRESHOLD_ = 0.01;
 JointTrajectoryAction::JointTrajectoryAction() :
     action_server_(node_, "joint_trajectory_action", boost::bind(&JointTrajectoryAction::goalCB, this, _1),
                    boost::bind(&JointTrajectoryAction::cancelCB, this, _1), false), has_active_goal_(false),
-                   controller_alive_(false), has_moved_once_(false), name_("joint_trajectory_action")
+                   controller_alive_(false), has_moved_once_(false), name_("joint_trajectory_action"),
+                   ignore_motion_server_error_(false), consider_status_unknowns_ok_(false)
```
simonschmeisser:
```suggestion
                   ignore_motion_server_error_(false), consider_status_unknowns_ok_(true)
```
It's a bit too strict like that

I didn't see the initialization here on first glance, doing it here or in the header is a mater of preference I guess

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -45,12 +45,28 @@ const double JointTrajectoryAction::DEFAULT_GOAL_THRESHOLD_ = 0.01;
...
 {
   ros::NodeHandle pn("~");
 
   pn.param("constraints/goal_threshold", goal_threshold_, DEFAULT_GOAL_THRESHOLD_);
 
+  // Two parameters for bw-compatibility with the 'old' behaviour.
+  // Setting these by default to TRUE, to maintain the previous behaviour
+  pn.param("ignore_motion_server_error", ignore_motion_server_error_, true);
```
simonschmeisser:
```suggestion
  pn.param("ignore_motion_server_error", ignore_motion_server_error_, ignore_motion_server_error_);
```
again a mater of preference I guess but it reduces the number of places where you can change the default

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -45,12 +45,28 @@ const double JointTrajectoryAction::DEFAULT_GOAL_THRESHOLD_ = 0.01;
...
+  pn.param("consider_status_unknowns_ok", consider_status_unknowns_ok_, true);
```
simonschmeisser:
```suggestion
  pn.param("consider_status_unknowns_ok", consider_status_unknowns_ok_, consider_status_unknowns_ok_);
```

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -123,6 +191,24 @@ void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle gh)
     return;
   }
 
+  // check robot can actually execute trajectory, if not, refuse goal.
+  // no point in accepting the goal only to cancel it immediately later
+  if (!is_motion_server_ok(last_robot_status_, consider_status_unknowns_ok_) && !ignore_motion_server_error_)
```
simonschmeisser: micro optimization ahead: check the bool first?

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -123,6 +191,24 @@ void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle gh)
...
+  {
+    // translate status into user readable description
+    const std::string reject_msg = {"Rejecting goal: "
+      + describe_robot_status_msg(last_robot_status_, consider_status_unknowns_ok_) };
+    ROS_ERROR_STREAM_NAMED(name_, reject_msg);
+    control_msgs::FollowJointTrajectoryResult rslt;
+    // the goal is actually probably OK, but we have to choose one of the existing
+    // constants, and this one comes closest
+    rslt.error_code = control_msgs::FollowJointTrajectoryResult::INVALID_GOAL;
```
simonschmeisser, line 203: we could also set the `rslt.error_string` ... not sure if anybody displays that though ... but I could imagine implementing that at some point in MoveIt

industrial_robot_client/launch/robot_interface_download.launch
gavanderhoorn: So this is indeed the main point of discussion here.

Doing this by default has the potential to break ppl's setups.

We could merge this PR as-is and not enable it on Melodic, then branch for Noetic and enable it there by default.

I hate branching for these little things, but I don't really see another option.

industrial_robot_client/src/joint_trajectory_action.cpp
gavanderhoorn: setting this to `true` does decrease the value of the checks.

Ideally this would be a per-field setting, but that was a bit of a bridge-too-far for me when I wrote this.

industrial_robot_client/src/joint_trajectory_action.cpp
gavanderhoorn: I think I'll leave it like this. It's indeed a very minor optimisation, and the current order reads almost like plain English: "if the motion server is not ok AND we don't ignore motion server errors, ...".

industrial_robot_client/src/joint_trajectory_action.cpp
gavanderhoorn: `setRejected(..)` does that for us: [api doc](https://docs.ros.org/en/melodic/api/actionlib/html/classactionlib_1_1ServerGoalHandle.html#af0a4d2debf5b19d985553903ddaef92c).

industrial_robot_client/src/joint_trajectory_action.cpp
simonschmeisser: Yes, it's a bit less strict that way but also less likely to actually break setups (?). Hard to know without looking up all the drivers out there

gavanderhoorn:
@simonschmeisser: thanks for the review.

Would you see an opportunity to test this on any real hw?

I'd be interested to know whether this mitigates what you reported in #131 for instance.

simonschmeisser:
I'll try to fit in testing that on our Mitsubishi in the evening

industrial_robot_client/launch/robot_interface_download.launch
simonschmeisser: If we leave it disabled, how would you enable it manually? copy paste the launch file?

I totally agree on avoiding branch-mess where possible

gavanderhoorn: I see now that I forgot to export the `param`s using `arg`s. I'll add that. That would allow overriding the defaults.

gavanderhoorn:
I actually don't have access to an *downloading* driver / robot, so I'd be interested to know whether the proposed changes actually work there.

If they do, we'll need to update the `.launch` file for download drivers as well.

simonschmeisser:
Thanks for pointing that out, I actually only have access to a _downloading_ robot/driver

gavanderhoorn:
> Thanks for pointing that out, I actually only have access to a _downloading_ robot/driver

I expect the changes to _accepting goals_ to work for both. I believe that should help with #131.

Aborting active goals will work, but whether the OEM server program also cooperates I can't say -- and will depend on its implementation of course.

gavanderhoorn:
Ok, so I've changed things around a bit:

 - defaults are only set once per context (in the JTA and as parameters in the `.launch` file)
 - by default everything is enabled: errors are not ignored, neither are UNKNOWNs treated as OK
 - `robot_interface_streaming.launch` exposes two new `arg`s, which allow changing the two bw-compatibility knobs introduced

We can still decide to set `consider_status_unknowns_ok` to `true` by default.

We could also say that the default should be `false`, and drivers which can't function that way should explicitly override the default `arg` and set it to `true`.

gavanderhoorn:
@simonschmeisser: any luck? :)

gavanderhoorn:
@simonschmeisser: *very* friendly ping.

simonschmeisser:
Sorry, I'm on the road (well tracks to be exact ...) but will try to give this some time this Friday

gavanderhoorn:
@simonschmeisser: friendly reminder.

(only doing this as I tend to forget when I sort-of promise someone I would do something three days later while sitting in a train ;) )

simonschmeisser:
The immediate feedback when being in manual mode/e-stop/drives not powered is quite nice.

```
[ERROR] [1625839551.379974859]: Rejecting goal: controller reported motion not possible (no further information)
[ WARN] [1625839551.380040937]: Controller  failed with error INVALID_GOAL: 
[ WARN] [1625839551.380065314]: Controller handle  reports status FAILED
[ INFO] [1625839551.380084931]: Completed trajectory execution with status FAILED ...
[ERROR] [1625839551.380121015]: Error executing plan:  "FAILED"
```
I wonder if this could or should be more specific?

Edit: e-stop is printed out correctly:
```
[ERROR] [1625840349.974565108]: Rejecting goal: controller reported e-stop

```

Otherwise I could not spot any regression with our mitsubishi downloading driver

simonschmeisser (dismissed):

gavanderhoorn:
> I wonder if this could or should be more specific?

well, we could inspect fields like `drives_powered` etc, but a) this was already enough work to implement and test (for a package basically in maintenance mode) and b) it gets quite confusing with drivers publishing `UNKNOWN`s.

Easy enough to do though: update `describeRobotStatusMsg(..)` in `industrial_robot_client/src/joint_trajectory_action.cpp` ([here](https://github.com/ros-industrial/industrial_core/pull/271/files#diff-ddfb1b2e4f9a12d814d443f1c6ecabbca5d682a6044344234c47b18e730776e5R137)).

Right now it singles out e-stops, then checks for any other errors (with potentially an OEM code) and finally it just checks `motion_possible`.

I'm not sure we can say for all supported systems that fi `drives_powered` is a requirement for `motion_possible` -- some robots have an "eco mode", which powers down the drives when they are not used for some time (even with a program running), and will automatically enable them again when needed.

These sort of things made me hesitant to check anything more than `motion_possible`, as at least for that field, we can rely on the "insight" and understanding of the driver developer to figure out whether motion would be possible and report it accordingly.

I would probably recommend applications (ie: other nodes) to subscribe to `robot_status` and inspect the messages if further information could help determine a course of action. The IRC doesn't have any way to communicate it, as we always only use `INVALID_GOAL`.

gavanderhoorn:
Just noticed this in your comment:

```
[ WARN] [1625839551.380040937]: Controller  failed with error INVALID_GOAL: 
```

I would have expected the message passed to `setRejected(..)` to show up after `INVALID_GOAL:`, but that doesn't appear to be the case.

To make the formatting for empty controller names somewhat easier to read, I've submitted https://github.com/ros-planning/moveit/pull/2761.

simonschmeisser:
Yes, fully agree.

One thing I was worried about is that this PR would reject goals in manual mode but it doesn't and that's a good decision as well. We show a confirmation dialog when in manual mode or anything else considered erroneous in our software and it can get annoying during initial bring up. So it's good that we don't need to change parameters to allow movements in manual mode.

gavanderhoorn:
> One thing I was worried about is that this PR would reject goals in manual mode but it doesn't and that's a good decision as well.

well, I didn't do anything special for that. It's all part of "only look at `motion_possible`".

If the driver reports motion should be possible, we're not going to doubt that by looking at other fields and then trying to be clever.

gavanderhoorn:
What we *could* do, is only include `drives_powered` et al. in the description (ie: the error message), but not in the check.

But I'm not sure that would work either, as it may lead to the wrong conclusion *on the human side* when they read the error message.

industrial_robot_client/include/industrial_robot_client/utils.h
```diff
@@ -36,6 +36,9 @@
 #include <string>
 #include <map>
 
+#include <industrial_msgs/RobotStatus.h>
```
gavanderhoorn:
```suggestion
#include <industrial_msgs/TriState.h>
```

simonschmeisser:
We would need to set the `error_text` string field of the `FollowJointTrajectory` msg for it to be printed there

I'm not sure why we have the empty controller name in all of our controller files ... what do you usually put there? Something like `mitsubishi driver`?

gavanderhoorn:
> We would need to set the `error_text` string field of the `FollowJointTrajectory` msg for it to be printed there

According to [the api doc](https://docs.ros.org/en/melodic/api/actionlib/html/classactionlib_1_1ServerGoalHandle.html#af0a4d2debf5b19d985553903ddaef92c) (and the sources), anything you pass to `setRejected(..)` gets assigned to the `error_text`. At least that's how I interpreted that.

> I'm not sure why we have the empty controller name in all of our controller files ... what do you usually put there? Something like `mitsubishi driver`?

It's most likely empty because the template [in the tutorial lots of people refer to](http://wiki.ros.org/Industrial/Tutorials/Create_a_MoveIt_Pkg_for_an_Industrial_Robot#Update_Configuration_Files) has it set to `""`.

If you set it to something non-empty, MoveIt will start looking in a different place for your action topics. That's not necessarily a problem though, you'd just have to update your driver to offer the action server there (or `remap`).

simonschmeisser:
> According to [the api doc](https://docs.ros.org/en/melodic/api/actionlib/html/classactionlib_1_1ServerGoalHandle.html#af0a4d2debf5b19d985553903ddaef92c) (and the sources), anything you pass to `setRejected(..)` gets assigned to the `error_text`. At least that's how I interpreted that.

But we have two text variables for the error here. The generic one you are setting and the redundant one in http://docs.ros.org/en/api/control_msgs/html/action/FollowJointTrajectory.html

gavanderhoorn:
O, crap, you're right.

I'd somehow confused those.

Perhaps MoveIt should check both and display them.

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -123,6 +189,24 @@ void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle gh)
...
+  if (!isMotionServerOK(last_robot_status_, consider_status_unknowns_ok_) && !ignore_motion_server_error_)
+  {
+    // translate status into user readable description
+    const std::string reject_msg = {"Rejecting goal: "
+      + describeRobotStatusMsg(last_robot_status_, consider_status_unknowns_ok_) };
+    ROS_ERROR_STREAM_NAMED(name_, reject_msg);
+    control_msgs::FollowJointTrajectoryResult rslt;
+    // the goal is actually probably OK, but we have to choose one of the existing
+    // constants, and this one comes closest
+    rslt.error_code = control_msgs::FollowJointTrajectoryResult::INVALID_GOAL;
```
gavanderhoorn, line 203:
```suggestion
    rslt.error_code = control_msgs::FollowJointTrajectoryResult::INVALID_GOAL;
    rslt.error_string = reject_msg;
```

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -227,9 +309,32 @@ void JointTrajectoryAction::controllerStateCB(const control_msgs::FollowJointTra
...
+      + describeRobotStatusMsg(last_robot_status_, consider_status_unknowns_ok_) };
+
+    // Stop the relay
+    stopRelay();
+
+    // return abort to action client
+    control_msgs::FollowJointTrajectoryResult result;
+    // would like to use a better error constant, but we have to choose one of the existing
+    // constants, and this one comes closest
+    result.error_code = control_msgs::FollowJointTrajectoryResult::INVALID_GOAL;
```
gavanderhoorn, line 329:
```suggestion
    result.error_code = control_msgs::FollowJointTrajectoryResult::INVALID_GOAL;
    result.error_string = abort_msg;
```

simonschmeisser (dismissed):

gavanderhoorn:
So with the technical side out of the way, the question becomes: do we want to merge this into `melodic-devel` and change the default behaviour?

@simonschmeisser @Levi-Armstrong @anyone-with-an-opinion-really?

gavanderhoorn:
With 0eeb23f0ff387a69deecaab66e9106e2f7af3228 and 0fa8bba32bbc1fc2a2e0bc140119ac4ab6e33817, MoveIt now reports something like:

```
[ WARN] [1626612647.088687582]: Controller '' failed with error INVALID_GOAL: Rejecting goal: controller reported motion not possible (no further information)
[ WARN] [1626612647.088809223]: Controller handle '' reports status FAILED
[ INFO] [1626612647.088865293]: Completed trajectory execution with status FAILED ...
[ INFO] [1626612647.088960767]: Execution completed: FAILED
```

The double *controller* there can be a bit confusing.

I was thinking of maybe adding "OEM" or "robot" there, so we would get:

```
[ WARN] [1626612647.088687582]: Controller '' failed with error INVALID_GOAL: Rejecting goal: OEM controller reported motion not possible (no further information)
```

But this would of course not be accurate if the server side would not run on an OEM controller.

simonschmeisser:
I'm not sure how many people understand what you mean by `OEM` so I'd rather go for `robot controller`

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -107,6 +121,58 @@ void JointTrajectoryAction::watchdog(const ros::TimerEvent &e)
...
+
+std::string describeRobotStatusMsg(industrial_msgs::RobotStatusConstPtr& msg, bool unknown_is_on = false)
+{
+  std::stringstream ss;
+
+  // mention e-stop specifically
+  if (utils::tri_state::isOn(msg->e_stopped, unknown_is_on))
+  {
+    ss.clear();
+    ss << "controller reported e-stop";
```
gavanderhoorn:
```suggestion
    ss << "robot controller reported e-stop";
```

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -107,6 +121,58 @@ void JointTrajectoryAction::watchdog(const ros::TimerEvent &e)
...
+  }
+  // some (generic ?) other error
+  else if (msg->error_code != 0 || utils::tri_state::isOn(msg->in_error, unknown_is_on))
+  {
+    ss.clear();
+    ss << "controller reported (active) error";
```
gavanderhoorn:
```suggestion
    ss << "robot controller reported (active) error";
```

industrial_robot_client/src/joint_trajectory_action.cpp
```diff
@@ -107,6 +121,58 @@ void JointTrajectoryAction::watchdog(const ros::TimerEvent &e)
...
+    }
+  }
+  // we use this last, as it could be we currently don't yet know *why*
+  // the state server decides motion_possible == false. So we first
+  // check the specific problems above, then fall back to this generic
+  // "it doesn't work, don't know why" statement
+  else if (utils::tri_state::isOff(msg->motion_possible, unknown_is_on))
+  {
+    ss.clear();
+    ss << "controller reported motion not possible (no further information)";
```
gavanderhoorn:
```suggestion
    ss << "robot controller reported motion not possible (no further information)";
```

simonschmeisser (dismissed):
There don't seem to be that many opinionated people around any more ...

I vote for merging this as is to `melodic-devel` as I see the slight change in behavior as a bug fix basically

industrial_robot_client/launch/robot_interface_download.launch
gavanderhoorn: Seems I've forgotten to fix the `arg`s missing. I'll need to address that before merging this.

simonschmeisser (approved):

gavanderhoorn:
Ok, to move this forward, I'm going to create a "post noetic" branch.

On *that* branch, everything discussed here will default to `false` (ie: `ignore_motion_server_error` and `consider_status_unknowns_ok`).

On both Melodic and Noetic, they'll default to `true` (to maintain the current behaviour).

People still using `industrial_robot_client` and likely to keep use it could build from source. The binary releases will just play-nice.

simonschmeisser:
Friendly ping, could be part of ros-o

PR #291, opened 2023-11-30
Fix clang -Werror complains

v4hn:
See the individual commits for the details.

- mark virtual overrides as override

- correctly define `CONNECT_HANDSHAKE` as unsigned

- fix bug with erase-remove idiom


v4hn:
@gavanderhoorn any chance you could merge so we don't have to carry more and more bugfixes [in the ros-o fork](https://github.com/ros-industrial/industrial_core/compare/melodic-devel...ros-o:industrial_core:obese-devel)?

PR #292, opened 2023-11-30
[ROS-O] No explicit requests for c++11

v4hn:
C++11 has been the default for compilers targeted in this branch.

gtest requires c++14 since v1.13 (Jan 2023) and building the tests on newer systems fails.

If this is not endorsed by the authors, it's also possible to check for externally provided values for `CXX_STANDARD` before setting it, but it's much more convenient to simply drop it and keep the codebase compiling without special triggers on modern systems.


v4hn:
Ping @gavanderhoorn . 🎅

